PR #45, opened 2024-04-02
Add Simplizied Chinese support

icyleaf:
Added Simplizied Chinese translate.


Flomp:
Thank you very much!
Adding a new language currently requires to change multiple files. I made the necessary changes and simplified Chinese is now available from v0.3.1 onwards.

PR #88, opened 2024-07-01
New language: Italian

lukasitaly:
I have completed the translation of all strings into Italian and incorporated the corresponding 'italian' string into each JSON file for the remaining languages.

### Two strings still need clarification: "about" and "off"

**About:** 
- If the context of _"about"_ is _"about 5km"_ then the correct translation is _"circa"_, 
- If the context is _"about"_ on the website as in _"about us"_ then the correct translation is _"su di noi"_

**"Off":** 
- If the context is _"turn off"_ then _"Spento"_ is the correct translation. 
- If the context is _"off by 5km"_, then translation would be different.

I hope the translations will prove useful and can be successfully implemented


Flomp:
Hi!
Thank you very much. Looks good!
Regarding your two strings:
"About" should be "su di noi" and "off" should be "Spento".

I'll include the translation in the next patch.

lukasitaly:
You're very welcome! Thanks you for all your work!

I changed the "about" string to "Su di noi". The "off" string was "Spento" since the first commit.

Issue #90, opened 2024-07-19
[BUG] Estimated duration can be 1h010Min


BastLast:
![image](https://github.com/user-attachments/assets/51f21f4f-3950-485f-8754-8b9c6b7397be)

Flomp:
Thanks, good spot. Merged it and will include it with the next patch.

PR #91, opened 2024-07-19
fix duration display #90

BastLast:


BastLast:
i did a small edit that should work, i have not tested it.

BastLast:
also maybe template string would be better for that kind of functions ?

Issue #128, opened 2024-11-15
garmin connect refuse to import a gpx file exported from wanderer


CPlusPlus17:
Import gives error, asked chatgpt about it:

_Yes, the XML snippet you provided is mostly valid, but there is a small issue with your xmlns attribute. Specifically, the xmlns attribute appears twice in the <gpx> tag, which is incorrect in XML. You only need one xmlns declaration for the gpx namespace. You can safely remove the duplicate one._

Modified

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<gpx version="1.1" creator="wanderer"
	xmlns="http://www.topografix.com/GPX/1/1"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd"
	xmlns="http://www.topografix.com/GPX/1/1">
	<metadata>
		<name>Biwak Test</name>
		<desc/>
		<time>2024-11-15T00:00:00Z</time>
		<keywords>jlrl2pyyj8hpjkh, </keywords>
		<author>
			<name>tlfrhfdi2gwfys7</name>
			<email>manuel.gysin@protonmail.com</email>
		</author>
	</metadata>
```

to

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<gpx version="1.1" creator="wanderer"
	xmlns="http://www.topografix.com/GPX/1/1"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.topografix.com/GPX/1/1 http://www.topografix.com/GPX/1/1/gpx.xsd">
	<metadata>
		<name>Biwak Test</name>
		<desc/>
		<time>2024-11-15T00:00:00Z</time>
		<keywords>jlrl2pyyj8hpjkh, </keywords>
		<author>
			<name>tlfrhfdi2gwfys7</name>
			<email>manuel.gysin@protonmail.com</email>
		</author>
	</metadata>
```

Import works now.

Flomp:
Now fixed in v0.12.0

CPlusPlus17:
Thanks you!

PR #131, opened 2024-11-23
Fix/gpx export chrome

tofublock:
This fixes a trail export problem with Chrome-based browsers.

The export function used to add a duplicate "xmlns" attribute to the <gpx> tag in Chrome and Edge, which led to a malformed GPX file (see #128). Additionally, GeoJSONs were produced as empty lists, like `{"type":"FeatureCollection","features":[]}`.

I fixed the malformed XML issue by declaring the default namespace and adding it back in for Firefox (which seems to swallow it), instead of the other way around. By doing that, the second problem resolved itself as well and GeoJSONs are now produced as expected.

Tested in Edge and Firefox. All outputs tested to be valid JSON and XML files.

### DIFF
--- a/web/src/lib/models/gpx/gpx.ts
+++ b/web/src/lib/models/gpx/gpx.ts
@@ -32,7 +32,7 @@ export default class GPX {
     $?: {
       version: string,
       creator: string,
-      'xmlns:prod': string,
+      xmlns: string,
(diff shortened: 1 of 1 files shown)


Flomp:
Looks good. Confirmed that it works. Thank you!

PR #138, opened 2024-12-05
Feature/trail table view

tofublock:
This adds a third ("table") view to /trails (could potentially replace the "list" view at some point?)
"Sort" dropdown is hidden in that view, as the column header elements are clickable.

Please let me know if this is in line with your style. If you have ideas how to improve table layout for responsiveness, I'm interested.


tofublock:
- I reused the sortOptions array from the TrailList component as the table header. If you dislike that let me know and I can decouple it.
- I rearranged the order of the sortOptions a bit so it makes more sense for the table, and I replaced "elevation loss" with "duration".
- I created a new migration for pocketbase to allow sorting by duration.

tofublock:
Added elevation loss back in and fixed the migration. Ready to merge if you're happy with it.

tofublock:
Last change: Add a badge if trail is set to public.
image

PR #213, opened 2025-03-23
Introduce new multi stage Docker image for DB component

Kami:
## Description

This pull request updates main `Dockerfile` to utilize multiple stages and build the binary inside the image as part of the build stage.

This should result in better developer experience and more reproducible builds.

If users want to use binary which has been pre-compiled locally outside the image (aka on host), they can do that `Dockerfile.pre_compiled` Dockerfile.

That one is almost the same as the original Dockerfile. I just added a check to fail in case the binary hasn't been correctly compiled locally (e.g. with `xgo` for `musl` or using `CGO_ENABLED=0`).

I also included a GHA workflow which builds the images to verify those files are working.

We could extend the original release workflow to build and push development version of images to Docker Hub / GHCR on each `main` branch run (we can use git commit sha as the image tag).

## TODO

- [ ] Get consensus on this change
- [ ] Update docs
- [ ] Verify Docker layer caching works
- [ ] Update release GHA workflow

### DIFF
--- a/.github/workflows/build_docker_images.yaml
+++ b/.github/workflows/build_docker_images.yaml
@@ -1,0 +1,80 @@
+name: Build Docker Images (DB)
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build_multistage:
+    name: Build Multi-Stage Docker Image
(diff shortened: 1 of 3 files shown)


cugu:
Closing this as it is outdated with the new docker files.

PR #216, opened 2025-03-23
Fix for double encryption of secrets

Kami:
## Description

Proposed workaround for a bug described https://github.com/Flomp/wanderer/issues/209.

WIP - more details and information to be added later.


Flomp:
LGTM. I'm fine with leaving the hacky way of detecting encrypted secrets for now. I think the performance implications are minimal.

PR #225, opened 2025-03-28
Ignore Strava activities with no GPX data

### LINKED ISSUE #186: Strava Not Importing More than 37 Activities
Hi, I used the Strava integration and the first time the sync ran it imported 37 activities and when I looked at my Strava developer account it looked like it had hit the API limit so I assumed it was due to that it didn't import anymore. However even waiting to the following day for the sync to run at 02:00 it didn't import anymore.

I then completely deleted all the data from the docker folder and re-created it and re-set up the Strava integration etc. The first import ran and imported 37 activities but again, even after waiting 24 hours it hadn't imported anymore.. I checked the Strava API and it had not hit any limits so it wasn't due to this.

I've logged into Pocketbase but I cannot see any logs at all at 02:00. I did try setting POCKETBASE_CRON_SYNC_SCHEDULE in the compose file and recreating the containers to force it to sync but again, it doesn't seem to attempt to import anymore than the 37 already in there.

In Strava I have about 130 activities.

Any ideas?

Cheers.

dyuri:
Ignores activities where `StartLatlng` is empty.
Fixes #186

### DIFF
--- a/db/integrations/strava/strava.go
+++ b/db/integrations/strava/strava.go
@@ -432,6 +432,10 @@ func fetchDetailedActivity(activity StravaActivity, accessToken string) (*Detail
 }
 
 func createTrailFromActivity(app core.App, activity *DetailedStravaActivity, gpx *filesystem.File, user string) error {
+	if len(activity.StartLatlng) < 2 {
+		return nil
+	}
+
 	collection, err := app.FindCollectionByNameOrId("trails")
 	if err != nil {
 		return err


Flomp:
Interesting, I never got to the function you changed. I manually created an empty activity with no GPX data. Then, fetching the GPX already failed in my case (`generateActivityGPX`). Do you have steps to create an activity in strava that your fix manages to import?

dyuri:
In my case, strength activities imported from Garmin caused the issue, like: https://www.strava.com/activities/13984376072

PR #228, opened 2025-03-29
Adding process-compose support for local development

stigi:
Hi,

Thanks again for wanderer.
I'm getting started on some unrelated contributions, but first I'd like to propose adding this process-compose file to the repo to make local dev much easier.

But a video probably speaks more than a million words:
https://github.com/user-attachments/assets/e8e30d23-d9e3-481c-a6a3-f72367a20ca8

Process-compose is a containerless docker-compose like tool for bringing up local processes.
Back in the Rails days, you would have tools like Procman or similar that offer a similar functionality, but process-compose has gained some popularity, especially in the Go community (from what I can tell, looking at it from the outside :))

Up for discussion:
- Should the tool also ensure the go toolchain is setup and installed?
- Should meilisearch be started in a docker context (process-compose is perfectly capable of starting docker jobs), instead of installed via homebrew

Also no hard feelings if you don't want to add additional tooling to the repo. I can keep this locally for my own usage as well.


cugu:
Closing this as we mostly focus on the docker-compose setup.

PR #285, opened 2025-04-30
Enhance Tags on TrailCard component with toggle support

briannelson95:
## Changes made
- modified tags on `trail_card.svelt` to show only first 2 tags by default
- added button after 2 tags to view more
- added function to allow for toggle of viewing more tags
- made text smaller in `chip.svelt`

## Why this is useful
- allows the `trail_card.svelt` to remain visual without taking away from the cover photo by using the tag options

### Before
Screenshot 2025-04-30 092259
### After
Screenshot 2025-04-30 092540

## Explanation
I like the use of adding tags to the trails but they take up a lot of space on the trail card view, taking away from the visual images of each trail. I made it so by default it will show the first 2 tags and show "+[x] more" which is a toggle allowing the user to expand the remaining tags if they choose to do so. I also made the text smaller "text-xs" on the `chip.svelt` to help with the visuals.

### DIFF
--- a/web/src/lib/components/base/chip.svelte
+++ b/web/src/lib/components/base/chip.svelte
@@ -16,7 +16,7 @@
         ? 'bg-primary text-white'
         : 'border border-input-border bg-menu-item-background-hover'} px-2 py-1 rounded-full flex items-center gap-1"
 >
-    <span class="text-sm">{text}</span>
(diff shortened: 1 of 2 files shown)


Flomp:
Thanks! I updated your PR to v0.17.0 and improved the UX a little. Good addition!

Issue #288, opened 2025-05-01
Uploading additional images removes previous uploaded images from track


garbast:
I've a track, where I uploaded two images both size 8 MB and 7 MB, saved the track and close the edit form.

I then edited the same track again and uploaded additional two images with 6,2MB and 9,3MB. Saving and closing the form I came back to the track view, in which the first two images are missing.

I don't think that this is what should happen. I'd expect to see all four images.

Flomp:
Should be fixed in v0.17.1. Feel free to reopen if not.

PR #294, opened 2025-05-08
[#288] fix uploading of photos to existing trail

slothful-vassal:
regarding to issue #288 I've tried to fix the uploading of new photos to an existing trail


Flomp:
LGTM. Thanks!

Issue #295, opened 2025-05-08
[Bug] Error when sharing a route


slothful-vassal:
Hi,

when I want to share a route, the selected user does not appear in the share modal dialog and in the database log the following error is shown:
"open templates/mail/notification.html: no such file or directory"

The same log entry appears, when adding a comment to a route of an other user.

But the curiosity is: the share/comment has been added succesfully. After refreshing the page, the share icon is visible and the other user can see the trail.

What could be the reason for this behaviour?

Many Thanks.

Flomp:
Thanks for your issue and PR. I found the same bug while working on the federation branch and fixed it before I saw your issue. It should work now with v0.17.0!

Issue #289, opened 2025-05-02
Adding comment results in error: Error posting comment


zmeel:
Hi! It's me again :-)

I really like the app so try using it but then I found that adding a comment to a trail does not work. I tested it also on the demo page and no luck.

Cheers,
Erik

slothful-vassal:
Hi,

can you please check the log entries on db side (pocketbase)? Maybe it is the same issue as I found in #295 

Thanks

Flomp:
Hey there,

I'm currently working on #215, which will completely change how comments are handled and hopefully also fix this issue. Please stay tuned.

slothful-vassal:
@zmeel Nevertheless, can you please check your db logs as mentioned above. Because if it is the same error as on my side, I don't think it will be solved by #215 (after a first quick code review).

zmeel:
Sorry, I did check the db logs but no mention of any error after trying to add a comment.

Flomp:
This is fixed with v0.17.0.

PR #264, opened 2025-04-18
Trail Table: Allow multi select

slothful-vassal:
Hello,

I've tried to implement a multiselect option for trail table view. There are still some open issues:
- missing 'Share'
- missing 'Export'
- horizontal scroll bar
- missing localizations

This "pull request" is a prototype to check out, if this is a feature, which complains with your plans to improve wanderer. If not, then I would stop here, before spending too much time ;)

Let me know, what you think about it.
Bildschirmfoto vom 2025-04-18 19-14-52

Many thanks.

#200

PR #296, opened 2025-05-08
fix retrieving notification mail template

slothful-vassal:
regarding to #295 and maybe (with luck) to #289


slothful-vassal:
solved with #264

Issue #215, opened 2025-03-23
feature request: Federation


lebochequirit:
Hi,
is there any chance to let different servers connect to each other, so that you can host your own one but still see all the other people's tracks?
ActivityPub might be the way.

Would be great.

Cheers
L.B.Q.R.

stigi:
I came here with the same question and found #196.

Not sure if @mfbehrens started looking into this already, but I would also be happy to chime in some of my dev-time, to explore a possible route towards a federated future.

mfbehrens:
I have not started anything yet. So feel free to do so, I currently got some other projects to work on.

Flomp:
I would love if wanderer could store data in a decentralized way. As mentioned in the other issue I'm completely unexperienced in implementing any such protocol. So please feel free to open a PR and we can start from there.

Flomp:
I started working on it..progress can be tracked here:
https://github.com/Flomp/wanderer/tree/federation

It will still take some time until I have a first beta. The changes are more involved than I thought.

I'll report back.

Flomp:
Please see #308.

stesie:
Hej 👋  I'm unsure whether this is the proper place to raise this (after all I'm super new to this project), but the federation/activity pub part sparked my interest ... and somehow makes me wonder.

And first of all, thank you for taking the time to add federation 👍 

Given that the federation code is based on the Activity Pub protocol, ... and that in the Fediverse usually it's possible to follow over applicaton boundaries, i.e. a Mastodon user may follow a PeerTube account ... do you consider this in or out of scope? Or put differently, should only Wanderer instances be capable of following other Wanderer instances, ... or shall a Mastodon user be able to follow a Wanderer account (and then maybe see the images and be able to comment, but not see the summit log)?

If the latter is not a goal anyways, kindly ignore the rest please.  Otherwise I wonder whether this is going in the right direction by introducing a custom/proprietary vocabulary, like e.g. the `type=trail` stuff. Or at least, to me, it feels like it's proprietary (the `@context` part seems to be missing also, so it's at least not pointing to some kind of official spec)

... next-best standard would IMHO be the Trip object that's defined on schema.org here -> https://schema.org/Trip

[...]

PR #307, opened 2025-05-22
Add waypoints to printed PDF

### LINKED ISSUE #304: Print route with waypoints
I store walking route instructions using waypoints, and I would like to be able to print these waypoints with the route, so that I have these instructions on the paper too. I don't think this is currently possible. I'm able to print the route description, but not the waypoint text. Ideally the waypoints would print with coordinates or a grid reference, and with the same icon as what's shown on the map for the waypoint, followed by the text they contain.

It might also be good if waypoints could have types (e.g. walking instructions or just photos etc), and then we could choose which types of waypoint to print (e.g. we probably only want to print walking instructions). But perhaps this feature is for another issue.

Would be great if this was implemented :)

RobertClarke64:
This PR adds a list of waypoints to the printed PDF of a trail. This doesn't currently print the waypoint images.

## Testing

I built and ran this locally using Docker. I printed without description or waypoints, with description only, with waypoints only, and with both enabled. These PDFs can be found below.

[description and waypoints.pdf](https://github.com/user-attachments/files/20401095/description.and.waypoints.pdf)
[description.pdf](https://github.com/user-attachments/files/20401096/description.pdf)
[waypoints.pdf](https://github.com/user-attachments/files/20401097/waypoints.pdf)
[neither description or waypoints.pdf](https://github.com/user-attachments/files/20401098/neither.description.or.waypoints.pdf)

Closes #304.

### DIFF
--- a/web/src/lib/i18n/locales/en.json
+++ b/web/src/lib/i18n/locales/en.json
@@ -160,6 +160,7 @@
   "import": "Import",
   "import-hint": "Select or drag GPX, FIT, KML or TCX files here...",
   "include-description": "Include description",
+  "include-waypoints": "Include waypoints",
   "integration-description-komoot": "Syncs your komoot tours with wanderer in regular intervals.",
   "integration-description-strava": "Syncs your strava routes & activities with wanderer in regular intervals.",
   "integration-disabled": "integration disabled",
--- a/web/src/routes/map/trail/[id]/print/+page.svelte
+++ b/web/src/routes/map/trail/[id]/print/+page.svelte
@@ -107,6 +107,15 @@
         return char.charCodeAt(0);
     }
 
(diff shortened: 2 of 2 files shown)


RobertClarke64:
Hey @Flomp, would you be able to take a look at this? Thanks :)

Flomp:
Hey man, 
Very sorry that I'm so unresponsive. I was very busy working on #215 and didn't want to create merge conflicts with the feature branch. So I put PR on hold. 
I'm hopefully mostly done with the implementation now and just finishing the documentation this week. So hopefully I'll be able to take a look at all PRs the week after.

Thanks for your patience 
Cheers
Flomp

RobertClarke64:
Btw @Flomp - I have just resolved conflicts here for your latest changes. Should be good to go again!

Flomp:
Looks good! Thank you very much!

PR #342, opened 2025-06-24
Update docker-compose.yml

pmffromspace:
use working meilisearch and newest images

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ x-common-env: &cenv
 services:
   search:
     container_name: wanderer-search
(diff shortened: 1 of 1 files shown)


Flomp:
Hi, 

Thanks for your PR. Upgrading meilisearch is a hassle:
https://www.meilisearch.com/docs/learn/update_and_migration/updating

You need to dump your old index and reimport it in the new version. This is why I keep ms updates to a minimum. Only if there is a security patch or a feature that is relevant to wanderer. For that reason I will leave the version pinned at v1.11.3 for now.

PR #386, opened 2025-07-09
Implement localization in calendar component

james-geiger:
This PR implements localization in the calendar component leveraging the svelte-i18n library.

1. For month name, generates a date from the `currentYear`, `currentMonth` and passes the date object to the library's `dateFormatter` using a custom `monthName` format defined in the library's init function.
2. For weekday name, implements a custom two-letter translation.


Flomp:
LGTM. Nice and elegeant 👍

Issue #426, opened 2025-07-30
Fresh installation cant login


salvah22:
Hi!

first of all thanks for the development of this app its really promising, sadly after a fresh install on v0.17.2 I cant login, im shown an error 500

<img width="1384" height="802" alt="Image" src="https://github.com/user-attachments/assets/a2f5795c-fcee-40ba-9568-7838344d757b" />

and the following stacktrace on javascript console

<img width="781" height="162" alt="Image" src="https://github.com/user-attachments/assets/96d6112d-5d0f-4816-aa18-de6c66131f54" />

cugu:
Does https://github.com/Flomp/wanderer/issues/384#issuecomment-3130074343  help?

salvah22:
@cugu thanks, the POCKETBASE_ENCRYPTION_KEY i set was indeed longer than 32 char,

i think that did the trick! days pulling me hair about this one haha

cheers

Issue #384, opened 2025-07-08
Unable to register a new user


plang115:
It's the same error as https://github.com/Flomp/wanderer/issues/14
When I want to register the first user I just get an error message at the bottom right that says that the user can't be created. There are no entries in the logs of the docker containers.
The origin url is right. I tried to access / register over https via my internal reverse proxy but I also tried to register over direct access to the ip+port without reverse proxy. Also tried different browser.
The docker compose project is running on a lubutunu 24.04 VM running on Proxmox 8.4.1. This is NOT an LXC, it's a VM. Docker Version on the VM 27.5.1.

My Docker Compose File:
`x-common-env: &cenv
  MEILI_URL: http://wanderer-meilisearch:7700
  MEILI_MASTER_KEY: vOLOsjPcfFANMANENepCHyDyGjscasDcdHnrb6X5pyXQPWo

services:
  wanderer-meilisearch:
    container_name: wanderer-meilisearch
    hostname: wanderer-meilisearch
    image: getmeili/meilisearch:v1.11.3
    environment:
      <<: *cenv
      MEILI_NO_ANALYTICS: true
    ports:
      - 7700:7700
    volumes:
      - /opt/docker/wanderer/meilisearch:/meili_data/data.ms
    restart: unless-stopped
    healthcheck:
      test: curl --fail http://localhost:7700/health || exit 1
      interval: 15s
      retries: 10
      start_period: 20s
      timeout: 10s
    labels:
      - com.centurylinklabs.watchtower.enable=true

  wanderer-db:
    container_name: wanderer-db
    hostname: wanderer-db
    image: flomp/wanderer-db:latest
    environment:
      <<: *cenv
      POCKETBASE_ENCRYPTION_KEY: wOLOsjPcfFANMANENepCHyDyGjscasDcdHnrb6X5pyXQPWo
      ORIGIN: My_Origin_URL (tried ip+port or fqdn over reverse proxy)
    ports:
      - "8090:8090"
    restart: unless-stopped
    volumes:
      - /opt/docker/Backup/wanderer/db:/pb_data
    labels:
      - com.centurylinklabs.watchtower.enable=true

  wanderer:
    container_name: wanderer
    hostname: wanderer
    image: flomp/wanderer-web:latest
    environment:
      <<: *cenv
      ORIGIN: My_Origin_URL (tried ip+port or fqdn over reverse proxy)
      BODY_SIZE_LIMIT: Infinity
[...]

PR #433, opened 2025-08-02
Verifiy settings more strictly

cugu:
- Verify `POCKETBASE_ENCRYPTION_KEY` is exactly 32 byte (prevent issues like #384, or #426)
- Verify `POCKETBASE_ENCRYPTION_KEY` is not the default one
- Exit db in case of any issues

This might actually break existing setup that use default keys 🤔

### DIFF
--- a/db/main.go
+++ b/db/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


aeble:
Maybe include an automatic generation on first run of the container?

cugu:
I'd rather create an install script to create the `docker-compose.yml` file. Creating the secrets at container start opens some questions on how to store and share them between containers.

aeble:
Could be written to a `.env` file in the data directory that is sourced?

Flomp:
I think using the default keys should not result in an error. If I just want to run wanderer in my private environment (or even publicly, without exposing pocketbase and meilisearch to the internet), this is not a problem.

If a user wants to change the keys, it should be a conscious, manual effort to make sure the user (hopefully) understands the implications of picking weak keys.

cugu:
@Flomp I moved most of the errors back to warnings.

I still exit for `len(os.Getenv("POCKETBASE_ENCRYPTION_KEY")) != 32`, as this will result in errors otherwise.

PR #440, opened 2025-08-03
fix: mapping of Komoot categories

### LINKED ISSUE #439: Bug in mapping Komoot categories
Some Komoot categories are not mapped correctly to according Wanderer categories, e.g. 
```
"mtb_easy":       "Workout",    // Gravelbike
"mtb_advanced":   "Walking",    // Enduro
```

Mik-:
Add missing category mappings and fix sports.
Also do exact mapping, if matching user defined categories are present.

Fixes #439

### DIFF
--- a/db/integrations/komoot/komoot.go
+++ b/db/integrations/komoot/komoot.go
@@ -221,6 +221,97 @@ func syncTrailWithTours(app core.App, k *KomootApi, i KomootIntegration, user st
 	return hasNewTours, nil
 }
 
+// mapCategory returns the category ID for the given Komoot category.
+// A two-step lookup will be done. First a lookup for a exact category will be performed. If the
+// user has manually defined the categories of the exactMap, these will be used.
+// If there is no result, a mapping to the default Wanderer categories is done.
+func mapCategory(app core.App, komootCategory string) string {
+	// List of sports retrieved from here: https://static.komoot.de/doc/external-api/v007/sports.html
+	exactMap := map[string]string{
+		// Main sport types
+		"hike":             "Hiking",
+		"mountaineering":   "Hiking",
+		"racebike":         "Racebiking",
+		"e_racebike":       "Racebiking",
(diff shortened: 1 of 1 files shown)


Mik-:
I do also a mapping to manually created categories, which fits better to the Komoot categories. A drawback is, that they must created exactly like defined in the code.

Maybe it's better to define a mapping table for this. And maybe this could be eligible to the Strava sync as well. But I think, the current version can help some guys and it would not break anything.

cugu:
Should the exact map map all values to unique values? Some people might want to tract e.g. E-MTB and MTB separatly.

Mik-:
The exact mapping is quit opinionated. There could/should be special E-Bike versions.
All tours, I've imported so far did not have a sport starting with `e_...`. I don't know how Komoot handle this.
I just took over all sports from the API docs. I also did not fiddle out how to create a tour on Komoot with sports marked as "Tracking sports type". Maybe this could be done via "Garmin watch to Komoot sync" or so.

As I'm pretty new Wanderer, I wanted to get the sync right with the default categories. IMHO the exact categories should be handled in a more user friendly way on the long run. But this is something, which concerns also other parts of the app. Nothing I want to propose as a newbie...

slothful-vassal:
Hello,

I like the idea of introducing a category mapping system. I don't know how @flomp feels about it, but I personally don't think your approach is flexible enough. I hope you don't mind, but I've tried implementing a different, more generic approach (see #649).

Best regards and thank you very much for your commitment.

slothful-vassal:
We’re currently moving the Komoot integration into a new plugin system. I compared the category list from this PR with the new Komoot plugin manifest in #1034, and the same provider category set is covered there. The specific Komoot sport labels are represented as plugin metadata, while the default mapping points to the built-in wanderer categories.

Since this is covered as part of #1034, I’m going to close this PR. Thanks again for your help!

Issue #428, opened 2025-07-31
Can't see OAuth providers


aeble:
I'm logged into the PocketBase backend but have no menu entry to configure OAuth2 providers. Do I need to set something up in the `docker-compose.yml` or somewhere else for these to show up?

cugu:
@aeble did you follow https://wanderer.to/run/backend-configuration/#enable-a-provider-in-pocketbase ?

aeble:
My bad, I apologize.

cugu:
> My bad, I apologize.

No need to apologise, and happy i could help!

PR #441, opened 2025-08-03
Change docs

cugu:
- Move SMPT and OAuth2 from Backend configuration to their own pages to give them more visibility (s. #428)
- Move changelog a level up
- Added redirects to still support old links


Flomp:
Good change 👍 
I just collapsed the new "Backend configuration" section by default and added some (hacky) CSS to make the nested collapse smaller than the outer ones.

PR #457, opened 2025-08-12
Trail Edit / Waypoints from photos: prevent creating different waypoints for near-by locations

slothful-vassal:
When creating waypoints based on a list of photos, a waypoint is generated for each photo, even if multiple photos were taken at the same location.
I tried combining photos taken within a small radius (50 m) into a single waypoint.
Perhaps the radius size should be configurable.


slothful-vassal:
as draft, because the radius has to be configurable first

slothful-vassal:
I've made the waypoint merge radius configurable in database - a new field added to category. Why category? Because the merge radius may differ between them. When hiking I'm slow and the waypoint maybe closer than when biking at higher speeds. The default is 50m.

Flomp:
Nice addition!
I can see the new database field `category->settings`. However, it is currently null. I also did not see any initialisation code in your PR. There should probably be a migration to set an initial value and some documentation on how to set the value for new custom categories.

Flomp:
Thanks, this works as advertised. 
I have one more wish: currently, this only works when I add all the photos in one go. Once I create waypoints from photos a second time, the waypoints are duplicated again (e.g. if I pick the same photos). I would expect that in this case, the selected photos are merged into the existing waypoints as before.

slothful-vassal (commented):
Thanks for the helpful suggestion! I’ve addressed it by:
- adding a merge dialog for nearby waypoints during manual waypoint creation
- attaching repeat photo imports to matching existing waypoints instead of creating duplicates

Flomp (approved):
Perfect! Great work!

Issue #430, opened 2025-08-01
Not possible to add more than 19 trails to a list


jornfranke:
I am using https://trails.tchncs.de/ and I have a list of trails, but I cannot add more than 19 trails. Is this some internal limitation or is it the server?
Wanderer version: v0.17.2

It fails with the 

POST https://trails.tchncs.de/api/v1/list/form/uh057zqgzj7m6lp

It returns a http status code of 400.

The response is
```
{"data":{},"message":"Something went wrong while processing your request.","status":400,"detail":{"data":{},"message":"Something went wrong while processing your request.","status":400}}
```

jornfranke:
jfy
It is a public list.
It seems when adding the tracks, it reuploads the tracks with all geocoordinates, which does not look like it makes sense.

slothful-vassal:
I can confirm that there is a limit, but not fix 19. It seems to depend on the total size.
You're right. Reason seems to be that always everything will be reuploaded. In the detailed log I found "message too large".

jornfranke:
Thanks a lot for confirming it.

Is there a reason to upload everything again? 

I would assume if the list is updated then only a list of ids of the trails is needed and not the trails themselves.

slothful-vassal:
No, there's no reason. I have excluded all trail details from upload now.

jornfranke:
Wow thank you a lot for the quick response.
I hope I will eventually find some time to get into Wanderer so that I can also contribute to it!

jornfranke:
Did this fix end up in 0.18.0? I am currently testing with 0.18.0  and if I click on "Save List" - nothing happens (I can observe also no network traffic). This is independent if I remove, add or do nothing with the trails in the list.

I just observe in the console
```
XML Parsing Error: syntax error
Location: https://trails.tchncs.de/lists/edit/uh057zqgzj7m6lp?avatar=&name=&q=
Line Number 1, Column 1:
```

slothful-vassal:
1. create new list -> works fine 
2. add new trails to existing list -> nothing happens
3. replace existing trails through other trails -> works fine
4. Keep the list open after 3. -> everything works fine

I hope, I can check this evening

slothful-vassal:
OK, found some minutes for further analysis, but no time to fix

- 0.17 + #460 -> OK
- 0.18 -> not OK

slothful-vassal:
I have fixed, but not sure if my solution will pass the quality check ;)

Flomp:
[...]

PR #460, opened 2025-08-15
prevent uploading track details when updating a list [#430]

slothful-vassal:
solves the problem mentioned in #430

took me 2 hours of analysis for one line of code... I like it ;)


Flomp:
Elegant fix. Thanks!

PR #469, opened 2025-08-23
Add issue templates

cugu:
Add issue templates for bug reports and feature requests. Redirect questions to discussions.

See https://github.com/cugu/issue_template/issues/new/choose for who this looks like finally.

❓ @Flomp: We could consider moving feature requests to [ideas](https://github.com/Flomp/wanderer/discussions/categories/ideas). Those display votes more prominently, but we might loose current "👍" votes.


Flomp:
Oh cool, I did not know that was possible on gh. Very good addition! 

Especially for the bug report, I think it would be a good idea to encourage users to upload their GPX track that caused the issue (if applicable). `.gpx` files are not allowed by gh, but you can simply circumvent it by changing the file type to `.txt`.

I agree with moving feature requests to the discussion section. Even if that means losing the reaction count.

cugu:
@Flomp awesome, I'll do that later and adapt this PR.

cugu:
@Flomp can this be merged?

Flomp:
Absolutely. I wasn't sure if you wanted to convert the feature requests first.

PR #471, opened 2025-08-23
Build docs image via GitHub actions

cugu:
Build and push docs image for every push on main.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -1,11 +1,15 @@
 name: Build Documentation
+
 on:
-    push:
-        branches:
-        - main
-    pull_request:
-        branches:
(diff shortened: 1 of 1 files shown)


Flomp:
The problem I see with running this on every push to main is that the docs will get out of sync with the latest version. Since the project is still fairly small I take the liberty of pushing directly to main. That means that changes that are not released will already end up in the docs.

cugu:
True, but also updates of the docs that do apply to the current version are not released until the next release.

But docs with versions are also terrible. Not sure on the best solution here.

BerendWouters:
Since you tag your releases @Flomp , wouldn't it be possible to build the docs whenever a new tag is pushed (to main)?

Issue #458, opened 2025-08-13
Respect nominatim.openstreetmap.org usage policy


edaigle:
Adding a new trail by uploading a GPX to the form on `/settings/export` fails (error message as in #412 although unsure if same root cause). After some digging, it fails because the backend gets a 403 from Nominatim when trying to geocode the area in `stores/search_store.ts:searchLocationReverse`:

```

const r = await fetch(`${nominatimURL}/reverse?lat=${lat}&lon=${lon}&format=geojson&addressdetails=1`, {
        method: "GET",
    });

....

Response {
  status: 403,
  statusText: 'Forbidden',
  headers: Headers {
    server: 'nginx',
    date: 'Wed, 13 Aug 2025 22:48:48 GMT',
    'content-type': 'text/html',
    'content-length': '941',
    connection: 'keep-alive',
    'keep-alive': 'timeout=20',
    etag: '"6731053f-3ad"'
  },
  body: ReadableStream { locked: false, state: 'readable', supportsBYOB: true },
  bodyUsed: false,
  ok: false,
  redirected: false,
  type: 'basic',
  url: 'https://nominatim.openstreetmap.org/reverse?lat=49.59505&lon=-121.12443&format=geojson&addressdetails=1'
}

```

This is because the app does not have a specific User-Agent as required by the API's [usage policy](https://operations.osmfoundation.org/policies/nominatim/). Adding a header fixes the issue:

```
const r = await fetch(`${nominatimURL}/reverse?lat=${lat}&lon=${lon}&format=geojson&addressdetails=1`, {
        method: "GET",
        headers: {"User-Agent": "My Fancy Wanderer User agent"}
    });
```

As best I can tell this doesn't happen when uploading a GPX to the individual New Trail UI because the request is then made on the frontend and inherits the browser user agent, which keeps Nominatim happy.

I can open a PR for this but assume there should first be discussion as to what the header should be.

slothful-vassal:
The main problem could be, that default installations only use this demo/test instance of nomimatim which seems not intended to be used in this way. To ensure to respect the usage policy in all aspects, e.g. wanderer has to log the number of calls and prevent sending more than allowed. I don't know, how nomimatim/osm logs the calls and if there is a black list. So at least (maybe), I would suggest to include the name of the wanderer instance in the user agent name, like 

`headers: { 'User-Agent': '${env.ORIGIN}/${version}' }`

Flomp:
[...]

PR #485, opened 2025-08-27
Add header to searchLocationReverse (#458)

edaigle:
As described in #458, uploading trails intermittently fails because the request to the Nominatim API in `searchLocationReverse` does not contain a header as required by the API. The header was added to `searchLocations` in 47f431e, but the issue with the trail upload was caused by it missing in `searchLocationReverse`. This PR adds it there as well.


Flomp:
Good catch, thanks!

PR #555, opened 2025-08-30
Performance improvements

slothful-vassal:
Hi,

on slower machines with large lists (resulting in large gpx data), there is a big delay when opening a list. The reason is fetchGpx. You can see this effect on your demo instance as well.

I have tried to skip fetching gpx detail when opening a list. Instead, gpx detail is fetched when needed (for the map).

Maybe you like, otherwise it would be nice, if you can make some improvements to reduce the mentioned delay.

Many thanks & regards.


slothful-vassal:
What have I done in the last few weeks?

- Refactored the map component to improve maintainability.
- Massively reduced the load times of map and list pages by:
    - Using trail bounding boxes to evaluate zoom.
    - Using polyines instead of GPX for lower zoom levels.
    - Using GPX files only for higher zoom levels.

I will update this PR after completing my tests in the next few days.

PR #563, opened 2025-09-03
fix opening profile/trails page from nav_bar

slothful-vassal:
Hi,

I have found an issue with open the profile/trails page from nav_bar (top right user icon). It's not working, when the username contains upper case letters.

What's the reason behind the 'toLowerCase'?

Thanks & Regards

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,6 @@ search/dumps
 
 run.sh
 build*.sh
-start.*
+start*.*
 
 data*/
--- a/web/src/lib/components/nav_bar.svelte
+++ b/web/src/lib/components/nav_bar.svelte
@@ -98,9 +98,9 @@
 
     function handleDropdownClick(item: { text: string; value: any }) {
         if (item.value == "profile") {
-            goto(`/profile/@${$currentUser?.username?.toLowerCase()}`);
+            goto(`/profile/@${$currentUser?.username}`);
         } else if (item.value == "trails") {
-            goto(`/profile/@${$currentUser?.username?.toLowerCase()}/trails`);
+            goto(`/profile/@${$currentUser?.username}/trails`);
         } else if (item.value == "logout") {
             logout();
             window.location.href = "/";


slothful-vassal:
@flomp: Any chance to integrate this fix in one of the next releases? Many thanks :)

Flomp:
Sorry, I didn't get back to you sooner. 
The current username situation is somewhat confusing.
There are three usernames:
1. username in the `users` table: used for logins and all other auth handling, but is not displayed anywhere. It can contain upper and lowercase letters
2. username in the `actvitiypub_actors` table: this is the same as the username in the `users` table, but is displayed e.g. in the profile page
3. preferred_username in the `actvitiypub_actors` table: this is the same as the username but strictly lowercase. It is used for all URLs that contain a handle. E.g.: `/profile/@<preferred_username>`

So casting user.toLowerCase() should result in a valid handle that can be resolved. This is the call that finally pulls the actor from the database:
https://github.com/Flomp/wanderer/blob/378f030b4ab8198f47fe40aa625925c3c4053ee0/db/federation/actor.go#L49

So this would only fail in a case where username.toLowerCase() !== preferred_username. Is that the case in your case?

slothful-vassal:
Hi, thanks.
Yes, Isn't this is the standard case if you a user with upper cases?

Flomp:
No, a user with an uppercase username should always have a purely lowercase preferred_username.

slothful-vassal:
Yes, you are right. But the profile page is not reachable with preferred_username. It's only reachable with username.

Flomp:
That's true, but as a I layed out `users.user.lower() == activitypub_actors.preferred_user`. This is the relevant function:
https://github.com/Flomp/wanderer/blob/149157604d599c8033e6eefb523bbb04567bd775/db/util/activitypub.go#L77

It is called everytime a new user is created:
https://github.com/Flomp/wanderer/blob/149157604d599c8033e6eefb523bbb04567bd775/db/main.go#L196

So is this not the case with your dataset? If so, then there must be a problem elsewhere in the code.

slothful-vassal:
Yes, anything is wrong. When I create a new user with uppercase, everything is fine. But my existing users with uppercase have this problem. Maybe a migration issue from pre-fediverse version?

slothful-vassal:
```
webfinger request failed: Get "https:///.well-known/webfinger?resource=acct:***@": http: no Host in request URL
```

Flomp:
> Yes, anything is wrong. When I create a new user with uppercase, everything is fine. But my existing users with uppercase have this problem. Maybe a migration issue from pre-fediverse version?

Can you check in pocketbase if your entry for this account in `activity_actors` is different from the ones that work?

slothful-vassal:
tomatoes on my eyes. already checked, but not seen :( yes, username and preferred_username are reversed :o what could be the reason for that?

Flomp:
In an early state of the implementation, I had it the wrong way around, because I didn't read the federation docs properly. So at some point I had to swap them with a migration: 
https://github.com/Flomp/wanderer/blob/main/db/migrations/1750774093_swap_preferred_username.go

So maybe this migration did not run properly in your case?

slothful-vassal:
OK, thank you :) I have fixed them manually.

PR #564, opened 2025-09-04
Fix oidc2 and oidc3 logos

wolffshots:
When looping through providers and getting images for the providers it uses the provider name to get the log which is correct for all logos except the OIDC (2) and OIDC (3) providers which use the same logo as OIDC (`oidc.svg`).
If you add the second or third option then the login page throws a 500 error because the image couldn't be found.

This just strips the numbers. Not the most elegant solution but it shouldn't break any of the other existing providers.


Flomp:
Good find. Thanks!

PR #571, opened 2025-09-09
new trail view options: create copy & change visibility

slothful-vassal:
Hi,

I tried to implement the feature request to allow publishing multiple trails with one single action #505.

There are some code changes that could probably have been handled better. Please review.

Many thanks & regards


slothful-vassal:
back to draft, because of missing error handling and success message

slothful-vassal:
option added to create a copy (variant) of an existing trail (as requested in #517)

Flomp:
Generally, two nice additions. 
For the bulk publishing, I changed a couple of things.

1. I reverted the change with the toggles in the dropdown. From a UX perspective, that seemed a bit weird
2. I now allow the publish/private for all trails that are your own. To decide what the bulk operation should be when selecting both public and private trails, I simply use a majority vote: >=50% public -> new Visibility: private.
3. I now wait for a trail update to be persisted in meilisearch before returning from pocketbase. Otherwise, the trails would update but not yet reflect the change the user just made.
4. Fixed a bug that caused the first page to load when performing bulk operations on any page other than 1
5. Adds a spinner for the Dropdown while bulk operations are performed.

---

Regarding the copying. I see that a trail is currently directly persisted to the DB, and then the user can edit it. I think it would be more elegant if the user is simply redirected to the edit page directly and the trail is already loaded in the background. Then the user can decide whether to save the trail manually as usual.

So the workflow would be.

Original trail -> Duplicate -> Fetch complete trail from DB -> redirect user to /edit/new -> prefill the form with the fetched trail -> user saves (or not).

I'll look into that tomorrow.

Sorry for the wall of text.

Flomp:
Ok, for the trail copying, I implemented what I outlined in my previous comment. The only limitation currently is duplicating the photos (both for the trail and waypoints), since the photo files are attached to the original trail, and we would need to somehow duplicate them on the client side. 

I also decided to allow duplicating trails by other users. This is essentially a shortcut for exporting the GPX and reimporting it.

Let me know what you think :)

slothful-vassal:
Hi,

thank you very much for spending time to improve my PR :) 

I like your changes, for the most part. I still have a few concerns about the terms used for duplicating and publishing. All other entries use a verb, but the new ones use nouns. This is slightly confusing.
- “Duplikat” in german -> is the selected trail a duplicate or is one created when it is selected? I would prefer “Duplizieren” in german.
- “Visibility: Public"-> is the selected trail public or can I publish it? Unfortunately, it's not that easy to improve this. “Publish” is still easy, but what appealing term can be used to make a trail ‘private’ again? “Hide” is maybe also misleading. Unfortunately, I'm at a loss here.

grafik

twaldecker:
Is it possible to copy a trail that was shared by another user or is public?

I am planning to create an account for everybody in my family. After uploading my gps track to the wanderer trail I want to copy that to all members that were with me in their account.

slothful-vassal:
> Is it possible to copy a trail that was shared by another user or is public?
> 
> I am planning to create an account for everybody in my family. After uploading my gps track to the wanderer trail I want to copy that to all members that were with me in their account.

Why duplicate the trails? I recommend just sharing your trail with your friends and family.

db/util/meilisearch.go
```diff
@@ -359,10 +359,18 @@ func UpdateTrail(app core.App, r *core.Record, author *core.Record, client meili
...
+	task, err := client.Index("trails").UpdateDocuments(documents)
+
+	if err != nil {
 		return err
 	}
 
+	interval := 500 * time.Millisecond
+	_, err = client.WaitForTask(task.TaskUID, interval)
+	if err != nil {
+		return fmt.Errorf("meilisearch update trail: error waiting for task completion: %v", err)
```
Flomp: Initially, this was log.Fatalf. Killing the whole application is a bit much, I think.

Flomp:
I changed the publish labels to "Set public/private". For German it is "Veröffentlichen/Verbergen". 
"Duplikat" is now "Duplizieren", which is, of course, correct.

Let me know if this can be merged.

Flomp (approved):

PR #577, opened 2025-09-13
select items per pages

slothful-vassal:
Hi,

added an option to define the number of search results shown per page (#576).

Regards.

### DIFF
--- a/web/src/lib/components/trail/trail_list.svelte
+++ b/web/src/lib/components/trail/trail_list.svelte
@@ -15,14 +15,14 @@
     interface Props {
         filter?: TrailFilter | null;
         trails: Trail[];
-        pagination?: { page: number; totalPages: number };
+        pagination?: { page: number; totalPages: number, items: number };
         loading?: boolean;
         fullWidthCards?: boolean;
         onupdate?: (
             filter: TrailFilter | null,
             selection: Set<Trail> | undefined,
         ) => void;
-        onpagination?: (page: number) => void;
+        onpagination?: (page: number, items?: number) => void;
     }
 
     let {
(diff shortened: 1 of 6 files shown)


Flomp:
I really appreciate all your work! And sorry that I have not merged any feature PRs yet. I want to get v18 fairly stable before I introduce new features. Please stay tuned.

slothful-vassal:
No problem. I have seen, in this PR is still a bug with 'show all'. I will fix this in the next days (hopefully).

slothful-vassal:
I still don't like the current implementation of "all". Because pagination of front layer is directly linked to api pagination which could lead to problems with huge datasets. What do you think?
I might split api calls in "all" case. I will try.

Flomp:
I think an "All" option does not really make sense. Meilisearch limits the max. number of results to [~1000](https://www.meilisearch.com/docs/learn/resources/known_limitations#maximum-number-of-results-per-search). And I think we don't need to change that for now. 

It would be great if the selected number of items per page could be saved to local storage (`localStorage.setItem("perPage", pagination.items);`) so that it's saved through a page reload.

Also, I think that base 10 increments are a bit more logical to the general user (10, 20, 30 etc.).

slothful-vassal:
Thanks again for your helpful feedback.
I didn't knew this Limit. In this case, I will remove the ALL option.
I will try to add the storage proposal too.
But I'm not sure about the increments. Because in desktop Browser, the cards view is optimized for multiples of 3. With decimal increments there is no option remaining which fills the last line. And the last line looks a bit strange if not complete. Maybe the increments can be dependent on view Mode. What do you prefer?

slothful-vassal:
I've changed the increments to 10, 25, 50, 100 (w/o 'all') for list and table view, and keep the 12, 24, 48, 96 (again w/o 'all') for cards view because of the 3 items per line layout.
Implemented usage of localStorage.

Flomp:
Works like a charm! Thank you very much!

PR #595, opened 2025-09-21
Add setup.sh script

### LINKED ISSUE #415: Meili Container doesn't start because of docker-compose.yml errors
### MEILI_NO_ANALYTICS
the current `docker-compose.yml` seems to be broken: when the `MEILI_NO_ANALYTICS` stanza is set to `true` the container doesn't start and terminates with the error 

```
[search] | error: invalid value 'True' for '--no-analytics'
[search] |   [possible values: true, false]
[search] |
[search] |   tip: a similar value exists: 'true'
[search] |
[search] | For more information, try '--help'.
```

I haven't debugged further where the capitalization occurs.

### MEILI_MASTER_KEY
For some reason the Meili container doesn't seem to honor the evironment variable in the `x-common-env` section in the `docker-compose.yml`: if it's only set there, the container starts up and logs a message stating that it is running without a master key.
If the same environment variable is added to the `environment:` section of the search service it starts up without any issue.

### Evironment

- arch linux
- podman 5.5.2

cugu:
Add a `setup.sh` script for a guided setup. The script creates a `docker-compose.yml` with secure keys and sets other relevant environment keys. 

Also closes #415


Flomp:
Very nice addition!
I changed a couple of things:
1. Move `setup.sh` to the static root of the docs so that the URL can be `https://wanderer.to/setup.sh`
2. Make `setup.sh` work on macOS systems
3. Change the wording in a couple of places

If you have no objections, I'm happy to merge it.

cugu:
@Flomp The changes are fine with me. 👍

PR #596, opened 2025-09-23
speed up POI retrieval by custom overpass api url

slothful-vassal:
Hi,

retrieving POIs as map layer is sometimes very, very slow (up to minutes). I figured out, that the used provider overpass.private.coffee is much slower than other providers such as overpass-api.de. 

This speed issue is reproducable on your demo instance.

I made the overpass api url configurable (public env variable).

Best regards.


Flomp:
Thanks. I just switched the default to overpass-api.de. It was significantly quicker, also in my tests.

PR #615, opened 2025-10-03
FIx html entities in waypoint descriptions in print PDF

RobertClarke64:
This PR fixes the display of HTML entities in waypoint descriptions in printed PDFs. It also corrects some of the line height logic.


Flomp:
Good catch. Thanks for the PR!

PR #629, opened 2025-10-19
Improve trail planning: option to re-order route anchors

slothful-vassal:
Hi,

I have tried to improve the trail planning capabilities by adding a drag-and-drop list of trail anchors, which allows the user to change the order of the route anchors.

I hope you like it!

grafik

Kind Regards.


slothful-vassal:
I have removed the 'snap to path' toggle again
- makes no sense (and was not working), if 'follow paths' is disabled
- should always be active, if 'follow paths' is enabled, which is the new behavior

wendelinism:
I was just going to request this feature. Great, it's being worked on!

slothful-vassal:
Back to draft, because of rate limitations by nominatim. We need a rate limiter for request against the nominatim demo instance, max 1 request / second is allowed. Before, this wasn't a big problem, but when retrieving locations for the anchors, we have too much traffic. This limiter has to detect whether the demo or an private instance is used.

slothful-vassal:
- rate limit for nominatim calls added
- request cache for nominatim calls added
- retry added for nominatim calls added (sometimes, I've got ERR_HTTP2_SERVER_REFUSED_STREAM)

slothful-vassal:
- fixed several issues
- fixed some speed issues

New Feature 1: option to add the location you searched for to the route (as trail anchor)
grafik

New Feature 2: option to use a POI (e.g. camp ground) as a route anchor
grafik

slothful-vassal:
state handling still not optimal

PR #628, opened 2025-10-17
hammerhead integration added

slothful-vassal:
Hi,

I've tried to add a new integration for Hammerhead Dashboard.

I hope you like it.

Kind Regards

PR #649, opened 2025-11-16
generic integration category mapping

slothful-vassal:
Hi,

inspired by #440, I tried to implement a more generic approach to map integration provider categories with wanderer trail categories.

Because I don't use Strava, a default mapping for Strava is still missing.
If #628 will be merged too, a default mapping for Hammerhead has to be added first.

Regards


slothful-vassal:
found the strava mapping in api documentation

slothful-vassal:
solved with #1034

PR #657, opened 2025-11-25
fix db container warnings (pocketbase)

slothful-vassal:
Hi,

Since upgrading to Pocketbase v0.30.0, the following database container log warning has appeared:

`WARN You are using modernc.org/sqlite v1.39.0 which differs from the expected and tested v1.38.2.`

As I dislike such warnings, I can think of two ways to eliminate them:
- Use the old SQLite version.
- Upgrade Pocketbase to the newest version.

I chose to upgrade Pocketbase.

Regards,


slothful-vassal:
new PR: #846

PR #666, opened 2025-12-15
narrow-wide trail list view mode added

slothful-vassal:
Hi,

Sorry for the PR spam.

I would like to enhance the trail table by adding a column chooser. To prepare this, I would like to introduce an option to use the full width of the screen.

Current situation (space is limited):
grafik

In the new wide view mode, there is much more space for further columns:
grafik

To switch the view, click on the button below. The setting is stored for each list mode individually.
grafik

regards,


Flomp:
Thanks. I made 3 suggestions:
1. Move the Button to the top right
2. Change the icons
3. Added i18n

Please let me know what you think.

slothful-vassal:
Hi, thanks :)

The new icon may be slightly too prominent. What do you think about using expand/compress instead?

image
image

versus

Bildschirmfoto vom 2026-03-26 18-38-03
Bildschirmfoto vom 2026-03-26 18-38-25

Flomp:
I like it 👍

Issue #424, opened 2025-07-29
When drawing new trail, settings for cycling seem not to be obeyed


peter76:
Hi, just started playing around with Wanderer, looks very nice, thanks!

I noticed though that when I want to draw a new trail and change it to cycling, none of the settings seem to be obeyed.
I did a quick inspection of the sent headers and there also it looks like the settings are missing, but I am no expert on this:-)

Hope I explained this clear enough.

Regards, Peter

cugu:
@peter76 does it work for you when you select "Cycling" first an then draw the route?

PR #694, opened 2026-01-04
Fix swapped options for auto and bicycle routing

hhornburg:
Fixes #548 / #424 (there was a simple mistake of using bicycle options when doing car routing and vice versa)


slothful-vassal (approved):
many thanks

PR #697, opened 2026-01-04
server sided external api calls

slothful-vassal:
Using your own instances of geocoding and routing engines is easier if the API calls are triggered on the server (only a certificate is needed on the server, not the client).


Flomp:
I think we should adhere to svelte nomenclature and not prefix private variables with private. So `PRIVATE_VALHALLA_URL->VALHALLA_URL`.

I'm also fine with dropping `PUBLIC_VALHALLA_ENABLED` altogether. We provide a working config in our default docker compose. If someone decides to mess with the `VALHALLA_URL` env var, I think it's enough if the API returns a meaningful error message. But we don't need to hide UI elements.

Flomp (approved):

Issue #687, opened 2026-01-01
Creating more than 29 lists causes problems editing routes and adding routes to lists


crazy-to-bike:
### What happened?

When more than 29 lists are created, it's not pssible to edit routes: the loading bar in the browser stops short before the end of loading and nothing happens.
Although it's not possiblle to add routes to a list, because the popup window with all lists to select doesn't appear.
Imho this also corresponds with [#682](https://github.com/Flomp/wanderer/issues/682)

Is it possible to change this behaviour?

### Version

v0.18.3

Guacam-Ole:
I am unsure if this is the same issue but I had on my machine completely on 100% CPU after a lot of lists (about 40) had been created by a user. It went so far i had to reboot because not even ssh worked anymore. (16 cores)

Can you verify that your CPU usage also went up on your machine,, @crazy-to-bike ?

(As a workaround I limited CPU in docker-compose to protect the rest of the machine)

crazy-to-bike:
Yes, CPU usage went up enormously. So imho this corresponds with #682

Guacam-Ole:
Still not sure if the same issue. But when creating a new list with about 8 tracks it seems to create 40 Lists instead of one. And all of them are broken (not displaying the tracks)

crazy-to-bike:
OK, I haven't such problems. The 29 lists I actually use with up to 97 tracks within work fine. Lists with many tracks only have a long loading time because the browser is loading the complete list instantly.
But when I create only one more list than 29, I can't add tracks to a list per bulk action and the CPU usage increases very high.

slothful-vassal:
I can confirm, when editing a trail, the fetch lists request produces a high cpu usage in pocketbase db and runs into a timeout. After the timeout, the webservice produces high cpu usage (node) without any chance to fix this - you have to restart the service. I will try to fix it in the next days.

Guacam-Ole:
(i just realized my issue is a bit different. Created a new ticket)

Thanks for your work @slothful-vassal !

slothful-vassal:
hopefully fixed with #701

PR #701, opened 2026-01-05
fix list fetch timeout

slothful-vassal:
Hi,

This PR should hopefully resolve the issue mentioned in #687.

The issue was that the pocketbase method getFullList was incompatible with perPage = -1. Removing the 'perPage' part of the search parameters resolved the issue in my test case.

Regards

### DIFF
--- a/web/src/lib/components/base/search.svelte
+++ b/web/src/lib/components/base/search.svelte
@@ -15,6 +15,7 @@
 
     interface Props {
         maxSearchLength?: number;
+        timeBetweenUpdates?: number,
         value?: string;
         items?: SearchItem[];
         placeholder?: string;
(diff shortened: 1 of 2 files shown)


slothful-vassal:
Sorry, worked on wrong branch :( The list search modal should be a separate PR. Should I split?

Information about list search modal:
The list select modal was not very helpful if you have more than 10 lists or so. I replaced it with a search modal, inspired by the share modal.

Old:
image

New:
image
image

Flomp (approved):
LGTM. I changed startsWith to includes so that you can search for substrings. I also parameterised the time between updates in the Search component. For local client-side searches like this one, it doesn't make sense to wait 500 ms between updates.

Issue #703, opened 2026-01-08
Tracks imported from Komoot are always private...


WorldTeacher:
### What happened?

... even if the user has set tracks to be public by default.

I am using wanderer in combination with komoot to get my tracks into adventurelog. 
After I created some tracks, I got reports that other people couldn't view the tracks, they get an error 404. To resolve this, I had to mark the trails as public, then I set my profile to have all tracks public. I assumed this would fix it, but upon adding new trails to adventurelog today, I noticed that the tracks from komoot were once again private.

### Version

v0.18.3

olealbers:
Same behavior when importing GPX files instead of using the Komoot Integration

Issue #681, opened 2025-12-31
Imported Tracks are always Private


Guacam-Ole:
### What happened?

I try to import a bunch of GPX files. While most of this works fine I have the issue that they are always marked as private, no matter what setting is there for the user.

It should always respect the Privacy settings just like creating a track manually

### Version

v0.17.0

PR #719, opened 2026-01-13
fix trail privacy

Flomp:
Fixes privacy issues reported in #703 #681


slothful-vassal:
Hi,

trails upload -> fine
integrations -> I would prefer further settings for each integration, like 'Privacy' with options 'keep privacy' and 'use profile setting'

Why? For me, it would be a bug if my Komoot trail visibility settings were lost during the import process. The visibility settings for new trails should only be applied when creating new trails in Wanderer, unless the user explicitly wants to apply them to integrations as well.

Flomp:
Good idea. Added.

PR #731, opened 2026-01-15
trail sharding (spatial + time buckets) to bypass Meilisearch’s 1000‑hit limit

slothful-vassal:
In order to circumvent the 1k-hit Meilisearch limit, I attempted to implement a cluster strategy. This is based on two dimensions:
- Time, for the trail list.
- Space, for the overview map.

Time: time buckets with a maximum of 1k trails are created and sorted by creation date. Each trail is assigned to a unique time bucket.

Space: The map (bbox) is divided into nodes. The entire globe is used as the start node. As soon as a node contains more than 1k trails, it is divided into four subnodes (usually, the start node is divided at -30/0 rather than 0/0, so the split runs through the Atlantic rather than Europe). A trail can be contained in several nodes if it crosses node boundaries.

All clusters have a global scope. Experiments with a per-user scope were unsuccessful (due to issues with migration speed, data handling and write speed, etc.). In my test cases involving 3k trails and five users, the global approach showed no disadvantages in terms of speed.


slothful-vassal:
- I wonder why I had to add a new API route. I'll see if it wouldn't be smarter to use the existing ones.
- env var has to be documented

PR #676, opened 2025-12-29
Improve container images/docker compose for self hosting

pstoeckle:
Hi there,

love the website/code! I added some security measures to the `Dockerfiles`/`docker-compose.yml` for the self-hosting.

PR #734, opened 2026-01-16
use gcc/cgo for db image

slothful-vassal:
Hi,

As mentioned in the official PocketBase documentation and blog posts, PocketBase/SQLite provides better performance with GCC (CGO_ENABLED). It's not a significant improvement, but it's worth using.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -62,6 +62,11 @@ jobs:
         with:
           go-version: '1.22'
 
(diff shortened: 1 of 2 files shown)


cugu:
I'd prefer to get https://github.com/open-wanderer/wanderer/pull/676 merged before. But there are some open action items.

cugu:
@slothful-vassal this is outdated now. I'll close this, feel free to open a new one.

PR #735, opened 2026-01-17
fix corrupt meilie index when bulk create trails with UI

slothful-vassal:
side-effect of 57b37d84c46d0114723bec004cae2f47a1e44b8e (#673)


slothful-vassal:
not reproducable anymore

PR #750, opened 2026-01-25
Return 401s when logged out of more activitypub api endpoints

RobertClarke64:
This PR updates a few more of the activitypub endpoints to return a 401 when requesting external users when the client is not logged in to Wanderer.

The reason for this is because lots of web crawlers are hitting my instance, causing it to constantly pull more external users into PocketBase, eventually slowing down the system significantly.

Some more testing of this by other people would be appreciated. Locally this seems to work fine, returning a 401 when logged out, and working as expected when logged in. However when I run this image on my production instance, I seem to get a 400 error when trying to visit external user pages when I'm logged in.

### DIFF
--- a/web/src/routes/api/v1/profile/[handle]/feed/+server.ts
+++ b/web/src/routes/api/v1/profile/[handle]/feed/+server.ts
@@ -1,6 +1,5 @@
 import { RecordListOptionsSchema } from '$lib/models/api/base_schema';
 import { type FeedItem } from '$lib/models/feed';
-import { splitUsername } from '$lib/util/activitypub_util';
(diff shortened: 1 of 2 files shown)


slothful-vassal:
Hi, 

Many thanks for this PR :+1: 

I have centralised the actor get call and reintroduced a more robust origin check for the handler. I hope this is OK with everyone.

RobertClarke64:
Thanks @slothful-vassal, looks good :)

RobertClarke64:
This is what I've been running on my server for a few days, and is behaving as I'd expect.

RobertClarke64:
Thanks for the review :). Anyone got merge perms?

PR #751, opened 2026-01-25
Add root CAs to db container

### LINKED ISSUE #752: No CA Certificates on DB Container
### What happened?

With the recent container image hardening, the containers no longer have any root CAs. As a result, pocketbase can no longer pull ActivityPub users from other sites.

### Version

main

RobertClarke64:
With the recent container image hardening, the containers no longer have any root CAs. As a result, pocketbase can no longer pull ActivityPub users from other sites.

This PR simply copies the SSL certs from the build container to the final image.

Fixes #752

### DIFF
--- a/db/Dockerfile
+++ b/db/Dockerfile
@@ -42,6 +42,7 @@ FROM scratch
 WORKDIR /
 
 COPY --from=build /app/pocketbase /pocketbase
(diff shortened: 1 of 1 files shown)


cugu:
Good catch, thanks!

PR #761, opened 2026-02-02
v0.18.4 Changelog

Flomp:
### Bug fixes
- Tags can now properly be removed from trails
- Creating more than 29 lists does no longer prevent lists from loading
- Fixes a bug that caused the GPS data to be removed from summit logs after editing
- Fixes an issue that caused an infinite loop when creating a list with federation being active

### Features
- Trails can now be added to multiple lists at once

### Translation
-  Adds Norwegian translation (thanks @palhaland)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,14 @@
 - Creating more than 29 lists does no longer prevent lists from loading
 - Fixes a bug that caused the GPS data to be removed from summit logs after editing
 - Fixes an issue that caused an infinite loop when creating a list with federation being active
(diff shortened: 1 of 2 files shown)


slothful-vassal:
maybe #654, #694, #719, #750 should be mentioned too?
- Fixed an issue where bicycle routing options were incorrectly applied to car routing, and vice versa. (#694)
- Fixed an issue where filtering by multiple categories did not work as expected. (#654)
- Fixed imported tracks being marked private despite public-by-default settings. (#719)
- ActivityPub: External user access now requires authentication (401). (#750)

Flomp:
Added

slothful-vassal (approved):

PR #785, opened 2026-02-06
fix trail upload for users with default privacy settings

slothful-vassal:
Hi,

For new users, the privacy settings are missing (the db field is empty). In this case, the trail upload from the settings page will fail.


Flomp:
I think settings can never be null, no? settings.privacy can, of course, be null.

PR #797, opened 2026-02-10
Fix vertical description spacing on printed page

RobertClarke64:
This PR fixes the vertical height calculation of the description in printed PDFs. The previous version didn't account for newlines properly, and the waypoints section would appear too high in some cases, covering part of the description.

## Testing

I ran this locally and printed a walk with a previously problematic description. The waypoints now print below the description instead of on top:

image


Flomp (approved):
Thank you very much. I tested it and it solves the issue.

PR #798, opened 2026-02-11
Fix installation link in README

Guacam-Ole:
Corrected the link for installation from source in README. (simply was a 404 before)


slothful-vassal:
Many thanks :)

PR #803, opened 2026-02-12
Docker compose with traefik as proxy

flusile:
This changes adds a folder with file to start wanderer together with a traefik setup.
Using traefik as a proxy has two benefits:

- it takes care for the https certificates
- it enables the host to server more services than just wanderer

For more details please look at README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ The first startup can take up to 90 seconds after which you can access the front
 
 > ⚠️ if you are using wanderer in a production environment make sure to change the MEILI_MASTER_KEY variable.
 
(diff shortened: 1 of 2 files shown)


slothful-vassal:
Hi, and thank you very much for your contribution — we really appreciate the effort you put into adding another installation method!

We recently opened a dedicated discussion thread to collect and share alternative installation approaches in one central place: #817 

For now, the core project officially supports Docker and installation from source. Additional installation methods are currently considered community-driven.

I'm going to close this PR for now, but we'd be very happy for you to document your solution in the discussion, so that others can discover and build on it.

Thanks again for your contribution!

Issue #680, opened 2025-12-30
the cron job for batch uploads ignores the filename


flusile:
### What happened?

the curl call for upload a track sets the form parameter name to the conten of an env-variable named base_name. But taht varoiable is never set.
I should be initialized as base_name=$(basename "$file") before.

### Version

v0.18.3

PR #834, opened 2026-02-21
Fix/cron.sh

flusile:
Fix for [issue 680](https://github.com/open-wanderer/wanderer/issues/680).

The filename now is given as name parameter to curl.

Additionally it is now able to run multiple times in parallel. I ran into this problem while uploading many files so the srcipt ran longer then 15 minutes.

I also reworked logging to be sure any log is using the same way.

In general I would prefer to have this script started from outside using a docker exec. Currently everybody provides CPU power every 15 minutes for nothing.


Flomp:
Thank you very much for your PR. It motivated me to have another look at the matter of auto-uploading. I decided to pivot away from a cron-based approach and use a file watcher system that immediately uploads files when they are added to the upload folder. This was implemented and merged via #886. 

Nonetheless, I want to thank you for your contribution!

Cheers
Flomp

PR #859, opened 2026-03-01
Fix go version in Dockerfile

cugu:
#851 updated `go.mod` to go 1.25 which does not align with the Go version in the dockerfile. This PR fixes this and adds dependabot config to automate the updates in the future.


slothful-vassal:
Actually all PR tests fail because of trouble with "go mod download". Is this issue related to this PR?

Issue #860, opened 2026-03-01
Mentioning fediverse user in comments does not work


docnero:
### What happened?

I could not mention a fediverse user in the comments within a route.

I looked a little bit into it, but am not sure. It seems if within the "activitypub_actors" the "username" differs from "preferred_username", it selects the "username" but should select the "preferred_username"

<img width="396" height="236" alt="Image" src="https://github.com/user-attachments/assets/906359b5-bdb7-4e87-aca9-fa70d915ff62" />

As hack, I manually change the "username" to the "preferred_username" and it worked.

Here the according snippet that is created and contains the wrong receiver nik b. instead of docnero.

`
  {
    "id": "sv246is4djz94o1",
    "created": "2026-03-01 12:35:22.974Z",
    "data": {
      "auth": "users",
      "details": "webfinger request failed: <nil>",
      "error": "Not found.",
      "execTime": 127.79645,
      "method": "GET",
      "referer": "",
      "remoteIP": "172.19.0.4",
      "status": 404,
      "type": "request",
      "url": "/activitypub/actor?resource=acct:%40nik%20b.%20%F0%9F%90%9D%40xxx&follows=true",
      "userAgent": "node",
      "userIP": "172.19.0.4"
    },
    "message": "GET /activitypub/actor?resource=acct:@nik b. 🐝@xxx&follows=true",
    "level": 8
  }`

### Version

v0.18.5

Flomp:
Fixed via #885

PR #885, opened 2026-03-11
Fix mentions when username != preferred_username

Flomp:
Fixes the issue described in #860.


Flomp:
Good catch! Fixed with latest commit

PR #886, opened 2026-03-11
Feature: watcher based bulk uploads

Flomp:
This PR proposes a new method of bulk uploading trails. Instead of running a cron bash script every 15 minutes to check for new trails, we run a file watcher sidecar alongside the main node process. This script monitors the user-defined `UPLOAD_FOLDER` and uploads new files directly when they are detected.

Furthermore, this makes use of the newly introduced API tokens. The watcher expects the file(s) to be in a subfolder whose name is a valid API token. This way, we can facilitate bulk uploading for multiple users.

Further testing and documentation are required.


web/Dockerfile
slothful-vassal, line 1: not tested, but changed path of curl may damage healthcheck?

web/watcher.js
```diff
@@ -0,0 +1,70 @@
...
+    } catch (err) {
+      console.error(`[File Watcher] Failed to create directory ${uploadPath}:`, err.message);
+    }
+  }
+
+  console.log(`[File Watcher] Service active. Watching: ${uploadPath}`);
+
+  const watcher = chokidar.watch(uploadPath, {
+    persistent: true,
+    ignoreInitial: true,
```
slothful-vassal, line 24: maybe, should be documented that files added when container is down, will be ignored?

web/Dockerfile
Flomp: You're absolutely right. Changed it. I'll make sure to mark it as a breaking change in the changelog.

web/watcher.js
Flomp: Yeah, changing ignoreInitial to false does not import files upon container launch for some reason. I don't think it's tragic. I added a note in the docs.

slothful-vassal (approved):

PR #920, opened 2026-04-15
Adds option to mark trail as completed

### LINKED ISSUE #698: New done routes from Komoot are suddenly marked as "not done" in Wanderer after Komoot-Integration-Sync
### What happened?

In Komoot there are 2 new routes in the "Done" section from 03.01.2026 and 04.01.2026. After a sync the routes are imported to wanderer, but they are marked as not done.
Imports until 01.01.2026 are marked correct.

### Version

v0.18.3

### LINKED ISSUE #914: Komoot import finished tracks
### What happened?

Hi
I just got my komoot data synced into a fresh wanderer instance and all my trails are marked as not terminated, even the ones that are terminated.

### Version

latest (not specified)

Flomp:
Upon popular request, this PR adds a toggle to mark a trail as completed. Summit Logs no longer determine whether a trail is completed or not.


slothful-vassal (commented):
Hi,

nice idea overall, but I still have a few findings:

- Do we need a migration for existing trails that already have summit log entries?
- When I add a summit log entry, the route still stays marked as "not completed". Maybe we should either update it automatically or ask the user?
- Very personal opinion: I’m not a big fan of the current icons. I’d suggest taking a look at `compass-drafting` and `flag-checkered` instead, but of course that’s up to you.

slothful-vassal (commented):
I also set the `completed` flag for Hammerhead activities.

Flomp:
Thanks. I implemented all three suggestions.
1. The migration happens right after the new column is added
2. There is now a pop-up if the added summit log is the first one on a not completed trail
3. Icons changed

slothful-vassal (commented):
Thank you for the updates. At first I was a bit confused during testing because the confirmation dialog did not appear, until I noticed that it is not implemented in the trail editor at all. Was that a deliberate decision?

slothful-vassal (approved):

PR #921, opened 2026-04-15
Improves multiselect

Flomp:
This is just a small UX fix. After selecting one trail from the overview (cards, list or table), clicking on other tracks does not open them but rather adds them to the selection. It just means that one does not need to tick the tiny checkbox for all the trails he wants to select.


slothful-vassal (commented):
Hi,

I like this PR, but there are two small UX issues:
- ctrl+click not working anymore
- context menu has changed -> open in new tab is missing

Flomp:
Fixed 👍

slothful-vassal (approved):

PR #929, opened 2026-04-17
Improve frontpage performance

Flomp:
This PR switches the static parts of the 3D model on the home page for a pre-rendered image. This should drastically improve performance for lower-end devices. Also, it improves the starry night sky in dark mode a little.

grafik


cugu (dismissed):

slothful-vassal (commented):
I’m a little confused because this change seems to make the welcome page noticeably slower than before. I’ll send you some screen recordings on Discord so you can take a look.

slothful-vassal (dismissed):

slothful-vassal (approved):

PR #930, opened 2026-04-17
Federation Refactoring & Architecture Improvements

Flomp:
### Overview
This PR represents a significant architectural refactoring of the backend to better support ActivityPub federation. The changes extract monolithic code from `main.go` into modular hooks and route handlers, while implementing federation support for trails, lists, comments, and user interactions. The PR tries to offload as much federation logic as possible from the frontend API to the go backend.

### Testing Recommendations
- [x] Test ActivityPub federation endpoints
- [x] Verify list/trail creation and federation across instances
- [x] Test waypoint creation with author tracking
- [x] Test remote trail/comment interactions
- [x] Verify integration endpoints (Strava, Komoot, Hammerhead)


Flomp:
GitHub CodeQL flags this because it sees a user-provided domain string being used to construct a URL for an outbound HTTP request. Historically, this allows Server-Side Request Forgery (SSRF), where an attacker could force our server to hit internal metadata services (like 169.254.169.254) or local services (localhost).

We have to allow users to contact arbitrary servers to allow federation. However, this PR mitigates this risk by using a custom util.SafeHTTPClient(). This client is "hardened" against SSRF in the following ways:

- Network-Level Validation (Dialer Hook): Instead of just checking the string URL, we intercept the request at the connection level (in DialContext).

- DNS Rebinding Protection: We resolve the domain and validate the IP address before the connection is made. We then dial the validated IP directly. This prevents "DNS Rebinding" attacks where an attacker switches a "safe" IP for a "malicious" one between the check and the use.

- Private IP Blocklist: We explicitly block all private, loopback, and reserved IP ranges (e.g., 10.x.x.x, 127.x.x.x, 169.254.x.x). Even if an attacker provides localhost or an internal IP, the SafeHTTPClient will refuse to connect.

- Outbound Rate Limiting: We’ve added a per-actor rate limiter keyed to the dbActor.Id. This ensures that even if a user tries to abuse the discovery process to spam a remote server, they are limited to 30 requests per minute.

slothful-vassal (commented):
Hi,

I ran into an issue while migrating existing waypoints:

"Error: failed to apply migration 1775998131_set_waypoint_authors.go: sql: no rows in result set"

It looks like the migration tries to read the old user field after it has already been removed in the previous waypoint migration, so existing waypoint authors can’t be backfilled correctly.

Flomp:
Ah, my bad. I fixed it by changing the migration order.

db/routes/activitypub.go
```diff
@@ -0,0 +1,192 @@
...
+	if origin == "" {
+		return fmt.Errorf("ORIGIN not set")
+	}
+
+	body, err := io.ReadAll(e.Request.Body)
+	if err != nil {
+		return err
+	}
+	var activity pub.Activity
+	activity.UnmarshalJSON(body)
```
slothful-vassal: no error handling?

db/federation/actor.go
```diff
@@ -293,29 +335,36 @@ func fetchRemoteActor(actor *core.Record, iri string, includeFollows bool) (*pub
...
 }
 
-func FetchCollection(actor *core.Record, url string) (*pub.OrderedCollection, error) {
+func FetchCollection(userActor *core.Record, collectionURL string) (*pub.OrderedCollection, error) {
 	encryptionKey := os.Getenv("POCKETBASE_ENCRYPTION_KEY")
 	if len(encryptionKey) == 0 {
 		return nil, fmt.Errorf("POCKETBASE_ENCRYPTION_KEY not set")
 	}
-	req, _ := http.NewRequest("GET", url, nil)
+	ctx := context.WithValue(context.Background(), "actorID", userActor.Id)
```
slothful-vassal: missing auth header will lead to '[PANIC RECOVER] runtime error: invalid memory address or nil pointer dereference goroutine 2740 [running]:
github.com/pocketbase/pocketbase/apis.NewRouter.panicRecover.func3.1()'

db/routes/remote_list.go
```diff
@@ -0,0 +1,190 @@
...
+		if f, err := downloadFile(origin, collection, remoteID, gpx); err == nil {
+			record.Set("avatar", f)
+		}
+	}
+}
+
+func syncTrails(txApp core.App, userActor *core.Record, list *core.Record, origin string, trails []any) error {
+	col, _ := txApp.FindCollectionByNameOrId("trails")
+
+	localTrails := make([]string, len(trails))
```
slothful-vassal: make([]string, 0, len(trails)) ?

db/federation/actor.go
```diff
@@ -165,11 +205,11 @@ func assembleActor(actor *core.Record, dbActor *core.Record, app core.App, inclu
 	} else {
 
 		// check if value is still cached
-		twoHoursAgo := time.Now().Add(-2 * time.Hour)
-		if !includeFollows && dbActor.GetDateTime("last_fetched").Time().After(twoHoursAgo) {
+		twoHoursAgo := time.Now().UTC().Add(-2 * time.Hour)
+		if dbActor.GetDateTime("last_fetched").Time().After(twoHoursAgo) {
```
slothful-vassal, line 207: check for includeFollows missing (was checked in main-branch)

db/routes/remote_list.go
```diff
@@ -0,0 +1,190 @@
...
+	return shell, nil
+}
+
+func performFullListSync(app core.App, userActor *core.Record, reqURL *url.URL, localList *core.Record) (*core.Record, error) {
+	iri := localList.GetString("iri")
+	remoteUrl, _ := url.Parse(iri)
+	remoteUrl.RawQuery = reqURL.RawQuery
+	origin := fmt.Sprintf("%s://%s", remoteUrl.Scheme, remoteUrl.Host)
+
+	res, err := http.Get(remoteUrl.String())
```
slothful-vassal: SafeHTTPClient ?

db/routes/remote_profile_follow.go
```diff
@@ -0,0 +1,155 @@
...
+		return e.NotFoundError("Actor not found", err)
+	}
+
+	collectionIRI := actor.GetString(followType)
+	if collectionIRI == "" {
+		return e.BadRequestError(fmt.Sprintf("Actor has no %s collection", followType), nil)
+	}
+
+	// 2. Fetch Remote Content
+	client := &http.Client{}
```
slothful-vassal: SafeHTTPClient ?

db/federation/actor.go
Flomp: Yes, I decided, that we can cache follows. I think it's ok if we show 2-hour-old follower/following counts.

db/federation/actor.go
Flomp: Do you have a way how I can reproduce this? By accessing a specific profile from the UI?

db/routes/activitypub.go
```diff
@@ -0,0 +1,192 @@
...
+	return e.JSON(http.StatusOK, map[string]any{"actor": actor, "error": nil})
+}
+
+func ActivitypubActivityProcess(e *core.RequestEvent) error {
+	origin := os.Getenv("ORIGIN")
+	if origin == "" {
+		return fmt.Errorf("ORIGIN not set")
+	}
+
+	body, err := io.ReadAll(e.Request.Body)
```
slothful-vassal, line 72: You're already using io.LimitReader in iriFromHandle — maybe it's worth applying the same pattern to the response body reads in the remote_* handlers?

db/federation/actor.go
slothful-vassal: Since this is a bit tricky to test end-to-end, I started first tests directly against the local instance with curl. From what I understand, actor lookups need to work without authentication?

`curl -s "http://localhost:8090/activitypub/actor?resource=@user@domain&follows=false"`

db/routes/activitypub.go
Flomp: The problem with the remote calls is that we have no way of knowing how long the content will be. A trail might have a very long description. Then the response would be cut off after however many bytes we choose to read.

db/federation/actor.go
Flomp: The opinions between fediverse apps differ here. Some apps require auth for every call. I think allowing non-authenticated actor lookups in wanderer is fine. In any case, it shouldn't crash the app.

db/routes/activitypub.go
slothful-vassal: You're right. Maybe we should add a timeout instead?

Flomp: I just added the SafeHttpClient for all activitypub-related HTTP calls. It comes with a 30s timeout by default.

db/routes/activitypub.go
Flomp: Added with 72df942.

db/routes/remote_list.go
Flomp: Added with 72df942.

db/routes/remote_profile_follow.go
Flomp: Added with 72df942.

db/routes/remote_list.go
Flomp: Added with 72df942.

slothful-vassal:
Hi,

My first findings after setting up two test instances. It might be worth checking this alongside #966 as well.

### Waypoints

Compared to `main`, this branch regresses the remote trail detail refresh path. The ActivityPub behavior is the same in both branches: `TrailFromActivity` returns early for already-known trails, and `ObjectFromTrail` does not include waypoints, so incoming `Update` activities won’t refresh newly added waypoints.

On `main`, however, opening a remote trail fetches the remote API directly and can show fresh `expand=waypoints_via_trail` data. In this branch, the new cached full-sync path only refreshes immediately for new/`needs_full_sync` trails, then falls back to a 60-minute background refresh, so newly added remote waypoints can remain invisible in the detail view.

### Trail Lists

I think this changes the remote list detail behavior compared to main. On main, opening a remote list fetches the remote API directly and returns the fresh expand.trails payload, so the list contents are visible even if the local cached list has no trail relations yet.

In this branch, remote lists are served from the new local sync/cache path. If the list was already created via ActivityPub, it has a local ID, so RemoteListGet skips performFullListSync and returns the cached list as-is. Since ListFromActivity does not include the contained trails, that cached list can be empty even though the remote list contains public trails.

There also seems to be a small copy/paste issue in the stale-cache path: after 60 minutes it calls performFullSync instead of performFullListSync, so the list contents may not be refreshed later either.

Flomp:
Thank you very much for your review! All very valid points that I addressed in the new commits. Here's the breakdown:

> Since this is a bit tricky to test end-to-end, I started first tests directly against the local instance with curl. From what I understand, actor lookups need to work without authentication?

The rate limiter did not handle anonymous users properly, as they don't have an actor ID for identification. Instead of passing an actor object down to functions that use SafeHTTPClient, we now require a context object. It can be acquired via [util.GetSafeActorContext](https://github.com/open-wanderer/wanderer/pull/930/changes#diff-5c34e2a595c48a7f7a8dcffece0880f9fd1c6c5e55a4b85b2a94f215ecf994d7R173). This function generates a unique identifier (actor:<actorId> for logged-in users, anon:<ip-address> otherwise) to be used in the rate limiter. To address privacy concerns, we hash this identifier with a rotating salt and also clear the rate limiter's in-memory map every minute. This way, an admin cannot reconstruct who tried to access which host. (f89537b + 09d1dae)

> My first findings after setting up two test instances. It might be worth checking this alongside https://github.com/open-wanderer/wanderer/pull/966 as well.

Fixed via #966. Thanks!

> Waypoints

You are correct regarding the way trails are synced/cached now. I opted for this more aggressive strategy to reduce the amount of network traffic wanderer creates, as I think it could be a problem for big instances that host popular trails.

`TrailFromActivity` now sets `needs_full_sync` to `true` if the record already exists. This way, the next time the user interacts with the trail waypoints/summit logs will be updated. (c5b9c2f)

> Trail Lists

I added a `needs_full_sync` column to the `lists` table. It works analogously to trails. (c5b9c2f)

> There also seems to be a small copy/paste issue in the stale-cache path: after 60 minutes it calls performFullSync instead of performFullListSync, so the list contents may not be refreshed later either.

Fixed with 9a85b7c.

slothful-vassal:
Thanks a lot for the fixes — almost there! Just one last finding.

### Duplicate comments for local users on remote trails

When a local user comments on a remote trail, their comment is stored without an iri. On the next page load, syncRemoteComments fetches comments from the remote instance and fails to deduplicate (lookup is by iri), resulting in the user seeing their own comment twice — once as a local author and once as a remote author.

Flomp:
Ah, sneaky. Good find. Comments and summit logs (same problem here) now also consider a possible local ID when looking for a duplicate:
https://github.com/open-wanderer/wanderer/pull/930/changes/d986c128e8a71f07fcd7e10f00093f6dfe08fdb6#diff-2042958bf4f7988583c3aa538cb7dfca062ddbfa622e786f34577c1f55dc8fdfR139

https://github.com/open-wanderer/wanderer/pull/930/changes/d986c128e8a71f07fcd7e10f00093f6dfe08fdb6#diff-9f991bdde92bc298c6bb9bce7278a0f76ca33b6d12e9e29e95904f5eb53af513R236

slothful-vassal (dismissed):

slothful-vassal (approved):

PR #931, opened 2026-04-18
Improved and updated Norwegian translation

palhaland:
Som places with plural translation had not been updated correctly, these are now handled. In addition the hammerhead integration added some new translations, and these are now translated.


Flomp:
Thank you very much. Generally, we prefer changes to translations via [crowdin](https://crowdin.com/project/wanderer).

Flomp (approved):

PR #935, opened 2026-04-18
Feature/meilisearch start point

### LINKED ISSUE #115: trail disappears if start point is not on map
as the title says... the trail disappears if I zoom in on an area and the start point is not in the zoomed in area.,

palhaland:
feat: improve map trail visibility with bounding box intersection

- Extracted and indexed full bounding boxes (min/max lat/lon) for all trails.
- Switched map search to use bounding box intersection filter, preventing tracks from disappearing when the start point is off-screen.
- Implemented aggressive zoom filtering to hide short tracks when zoomed out.
- Added 'Map Cluster Zoom Level' configuration setting to allow user-defined clustering thresholds.
- Added data migration to automatically re-index existing trails with bounding box data.
- Added translations for the new setting across all 14 supported languages.

Fixes #115 

chore: upgrade dependencies and optimize docker builds

- Upgraded PocketBase to v0.36.9 and updated Meilisearch JS client.
- Resolved 11 security vulnerabilities in web dependencies.
- Added package overrides for secure versions of cookie, xml2js, and esbuild.
- Optimized Dockerfiles with better build flags and stage management.
- Fixed Meilisearch SDK capitalization breaking changes.

Programming by Gemini CLI.


palhaland:
@Flomp and @slothful-vassal , you might want to have a look. The chore (upgrade dependencies and optimize docker build) is strictly not needed (thus i separated them as two commits). But when building the docker for testing the npm gave some warnings about 6 medium vulnerabilities, and 5 other issues. Asked Gemini to fix it and also upgrade other dependencies.

web/static/docs/api/wanderer.openapi.json
palhaland, line 1: This could potentially be reverted

slothful-vassal:
Thank you very much for taking the time to submit this pull request. We appreciate your contribution and look forward to reviewing it more closely. We noticed that there seems to be some overlap with PRs #731, #555. Generally, we encourage keeping unrelated changes in separate pull requests, such as separating a new feature from a security fix.

db/migrations/1776510000_update_meilisearch_bounds.go
```diff
@@ -0,0 +1,55 @@
...
+				All(&trails)
+			if err != nil {
+				return err
+			}
+			if len(trails) == 0 {
+				break
+			}
+
+			if err := util.IndexTrails(app, trails, client); err != nil {
+				continue
```
slothful-vassal: Could this result in an endless loop here, since the page number isn’t incremented?

slothful-vassal (changes requested):
At the configured zoom level, the trails are no longer visible on the map. Neither the polyline nor the cluster is rendered.

db/migrations/1776510000_update_meilisearch_bounds.go
palhaland: That is indeed a good observation.

The error from IndexTrails may happen if:
1. Meilisearch is down or there are communication errors (new key)
2. The data size is related to 100 batches might exceed the meilisearch max_payload_size.

should there be added some logic to handle the max_payload_size and return an error in case of case number 1?

palhaland:
> At the configured zoom level, the trails are no longer visible on the map. Neither the polyline nor the cluster is rendered.

This is a good comment, and that may be because of the filtering setup when zooming out. What I set in the code is that routes shorter than 25 km could be removed at a certain distance zoom, but this should probably be user configurable.

slothful-vassal:
Thank you very much for this great contribution - we really appreciate it!

However, before we continue with this pull request, could you please do us a small favor? Could you split the independent changes into separate pull requests?

The following changes appear to be clean, self-contained adjustments that are easier to review and can be safely reverted in isolation.

- Meilisearch SDK rename + dependency upgrades
- Layer manager bug fix
- Error handling fallback in the search endpoints
- Dockerfile build optimization
- GeoJSON Polygon support
- Null safety fixes


This separation ensures a clear Git history and makes debugging easier should unexpected issues arise later.

I have a few minor suggestions regarding the main feature itself. Separate comments on this will follow once the separation is complete.

I hope this is okay with you.

palhaland:
Closed as this will be split into separate pull requests

palhaland:
@slothful-vassal , you can now provide comments on #963

PR #942, opened 2026-04-22
fix hillshading visibility

### LINKED ISSUE #747: Hillshading on 2D map
### What happened?

I'm using wanderer with its 3D terrain feature using [mapterhorn](https://github.com/mapterhorn/mapterhorn) as terrain tile provider. 

Configuration in `Settings -> Map`:
- _Terrain URL_: https://tiles.mapterhorn.com/tilejson.json
- _Hillshading URL_: https://tiles.mapterhorn.com/tilejson.json (does not make any difference whether this is set or empty, hill-shading looks identical either way with the same issue)

This mostly works, i can toggle between 2D and 3D map and can rotate the 3D view like expected.

What doesn't work is disabling the 3D view completely. Toggling to 2D removes any 3 dimensional mountains but the hillshading stays the same, this makes the 2D map less usable as the shading is making the map harder to read.

Only completely removing the URL from the Map-Settings removes the hill-shading on the 2D map.

I tested this on Firefox and Brave with the same results.

Thanks for the really nice project and your time :)

Edit: They seem to server hill-shading and terrain tiles from the same URL as per their documentation: [mapterhorn.com](https://mapterhorn.com/examples/terrain/)

### Version

v0.18.3

slothful-vassal:
fixes #747, if this is not the desired behaviour

### DIFF
--- a/web/src/lib/components/trail/map_with_elevation_maplibre.svelte
+++ b/web/src/lib/components/trail/map_with_elevation_maplibre.svelte
@@ -115,6 +115,7 @@
     let hoveringTrail: boolean = false;
 
     let mapLoaded: boolean = false;
+    let terrainEnabled: boolean | null = null;
 
     const trailColors = [
         "#3549bb", // blue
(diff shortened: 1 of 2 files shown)


Flomp (approved):
Fix confirmed

PR #950, opened 2026-04-28
Fix healthcheck command in docker-compose.yml

mfortini:
curl is not in the path, /curl is

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -76,7 +76,7 @@ services:
       - wanderer
     restart: unless-stopped
     healthcheck:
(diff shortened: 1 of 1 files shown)


Flomp:
The file you see is for the latest development version (the "main branch"). It includes changes that haven't been officially released yet.

If you are using the official pre-made Docker images, you need a version of the file that matches your image.

- Don't use the file from the main branch.
- Do go to the official repository, click on the "Tags" or "Releases" dropdown, and select the version number that matches the Docker image you downloaded. E.g.: [v0.18.5](https://github.com/open-wanderer/wanderer/blob/v0.18.5/docker-compose.yml)

PR #957, opened 2026-05-05
Fix syntax error in query selector for color scheme meta tag

mfortini:


slothful-vassal (approved):
Thanks for the fix — the selector syntax is indeed incorrect and the correction is appreciated! However, we couldn't reproduce any actual error in Brave/Chrome or Firefox. They seem to handle the missing ] gracefully and still find the element correctly. Could you let us know which browser you observed the issue in?

mfortini:
I actually was looking into another issue while I found that typo. Still have to investigate the original one :) Thank you

PR #958, opened 2026-05-05
meilisearch sdk rename and dependency upgrades

palhaland:
Pull request number 1: required for other pull requests, this is a follow up on #935


slothful-vassal (changes requested):
Hey, thanks for the meilisearch upgrade! One small note on the Dockerfile change: replacing npm ci with npm install works, but npm ci is generally preferred in build environments because it enforces that package-lock.json and package.json are in sync, giving you reproducible and deterministic installs.

If npm ci was failing locally, it's likely because the lockfile was regenerated with a different npm version and has a slight inconsistency with package.json. Could you double-check by running npm ci locally and committing a fresh package-lock.json if needed? That way we can revert to npm ci in the Dockerfile and keep the build deterministic.

palhaland:
you are quite correct, forgot to change it back after updating the npm packages to latest version, will revert the change to "npm ci"

slothful-vassal (approved):

slothful-vassal:
@Flomp OK to merge this PR? Would affect #930 only with a tiny change.

PR #959, opened 2026-05-05
dockerfile build optimizations

palhaland:
Related to #935 , and follow up of #958


slothful-vassal (changes requested):
Thanks for the db docker image optimization! 

Looks like the db/main.go and web/Dockerfile changes snuck in here — could you remove them from this PR since they're already covered separately?

slothful-vassal (approved):

PR #960, opened 2026-05-05
fix: layer manager bug and null safety fixes

palhaland:


slothful-vassal (dismissed):
Hey, thanks for the fix in maplibre-layer-manager.ts — clean and minimal, exactly what's needed!

That said, could you please remove the unrelated changes to db/main.go and web/Dockerfile from this PR? Those belong in their own PRs to keep things focused and easy to review.

slothful-vassal (approved):

PR #962, opened 2026-05-05
feat: geojson polygon support and gpx improvements

palhaland:


slothful-vassal:
Thanks for the contribution! Before we merge this, could you share the specific use case you had in mind? A sample file that triggers the new polygon behavior would also be really helpful — it would let us verify the conversion is correct and think through edge cases like KML files that contain both a polygon and a line string.

palhaland:
It was a good point testing it with a sample. 

The idea was to use it to mark a site or an area. For instance when hiking it would be nice ta mark areas that are good for camping/parking. Or when you encounter a place where the area is marked as "Private Property". 

When looking at how the polygon is rendered, it does not look like how I envisioned it, and there may not be that much use for it. We can close this for now. :)

PR #963, opened 2026-05-05
feat: meilisearch bounding box intersection and map filtering

palhaland:

Feature Overview: Map Clustering & Zoom Optimization

  To address performance and prevent the map from being overwhelmed with thousands of tracks at low zoom levels,
  this branch introduces a tiered clustering and visibility system. It balances server performance, network traffic,
  and user preference.

  1. The Bounding Box Diagonal (bounding_box_diagonal)
  Instead of filtering tracks by their actual route distance (which might zig-zag inside a small area), we calculate
  the diagonal length of the track's bounding box.
   * Why? This accurately represents the visual size of the track on the map. A 50km track tightly wound inside a
     2km forest will look like a single speck when zoomed out, whereas a 50km point-to-point track will span across
     the screen.
   * By filtering on the diagonal, we efficiently determine which tracks are visually relevant at any given
     altitude.

  2. Tiered Visibility & Zooming Out (Performance)
  The system uses environment variables to define "tiers" of visibility based on the current zoom level and the
  track's diagonal. This is critical for performance when the user zooms out and pans across large areas.

  Instead of forcing the browser to maintain and redraw thousands of tiny, sub-pixel tracks, the map dynamically
  sheds detail as you zoom out:
   * High Zoom (Detailed View): Shows all polylines or individual markers.
   * Medium Zoom (PUBLIC_MAP_MEDIUM_ZOOM_THRESHOLD defaults to 10): As you zoom out past this threshold, the map
     drops all small tracks (diagonal < 10000 meters) from the unclustered rendering queue.
   * Low Zoom (PUBLIC_MAP_LOW_ZOOM_THRESHOLD defaults to 8): Zooming out further drops medium tracks (diagonal <
     25000 meters).

  The Performance Impact: When a user is zoomed out to a country level and starts panning, the browser is only
  calculating draw-calls for a handful of massive, visually significant tracks, rather than choking on rendering
  10,000 tiny local hikes. (Note: These hidden tracks are still accurately summarized in the cluster bubbles—no data
  is "lost" to the user).

  3. User Preference vs. System Limits
  While environment variables protect the server and browser from overload, the user has ultimate control over when
  to transition from "Clusters/Dots" to actual "Polylines".

   * settings.behavior.mapClusterMinZoom: This user setting dictates the exact zoom level where clusters disappear
     and the full trail paths are drawn.
   * If the user has a powerful device and wants to see polylines earlier, they can set this to a lower number
     (e.g., 5). The system perfectly syncs the data fetching and map rendering layers to this single threshold,
     eliminating "dead zones" or "rough transitions" where tracks overlap or disappear.
   * If the user hasn't configured this, it defaults to the PUBLIC_MAP_HIGH_ZOOM_THRESHOLD (default 12).

  4. Two-Tiered Network Optimization
  To ensure the cluster counts are 100% accurate at global zoom levels without destroying network bandwidth, we
  implemented a dual-query approach:
   1. Summary Query: Fetches extremely lightweight objects (id, _geo, bounding_box_diagonal) for all tracks in the
      viewport. This guarantees accurate cluster numbers regardless of how far out you zoom.
   2. Detailed Query: Fetches the heavy metadata (Polylines, Names, Authors, Tags) only for the tracks that pass the
      diagonal visibility filter at the current zoom level.
   3. Client-Side Cache: Once a detailed track is downloaded, the frontend caches it. Panning back to a previously
      viewed area requires 0 additional network traffic for detailed metadata.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Additional CSRF/SSRF protections and rate limiting have been implemented for ActivityPub and outbound network calls. (PR #930)
 
 ## Features
+- Tiered map clustering and polyline filtering: The world map now uses zoom-dependent thresholds to manage performance while maintaining transparency. Small trails are automatically clustered at lower zoom levels, while larger routes are shown as detailed polylines. These thresholds and visibility limits are fully configurable via new environment variables (`PUBLIC_MAP_*_ZOOM_THRESHOLD` and `PUBLIC_MAP_*_DIAGONAL_LIMIT`).
 - Hammerhead integration added, including synchronisation of planned and completed tours, and manual trail sending. (PR #628)
 - The federation has been significantly expanded and refactored to provide more robust remote content synchronisation and local caching for remote trails and lists. (PR #930)
 - Trails can now be explicitly marked as completed. (PR #920)
@@ -50,7 +51,6 @@
 
 ## Maintenance
 - Meilisearch, PocketBase, Go, web/docs dependencies, CI actions, and Docker build setup updated.
-
 
 # v0.18.5
 ## Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# [Unreleased]
+
+## Features
+- Tiered map clustering and polyline filtering: The world map now uses zoom-dependent thresholds to manage performance while maintaining transparency. Small trails are automatically clustered at lower zoom levels, while larger routes are shown as detailed polylines. These thresholds and visibility limits are fully configurable via new environment variables (`PUBLIC_MAP_*_ZOOM_THRESHOLD` and `PUBLIC_MAP_*_DIAGONAL_LIMIT`).
+
 # v0.19.0
 
 ## Breaking Changes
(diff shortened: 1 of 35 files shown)


slothful-vassal:
Thank you very much for this contribution. The bounding box intersection approach is a significant improvement over the previous centroid-based filtering method. Separating the clustering data from the polyline data is an elegant solution. I have a few questions that came up during the review process.

I must admit that I don't fully understand all the interactions between the environment variables and the user settings yet. However, something feels slightly off. PUBLIC_MAP_HIGH_ZOOM_THRESHOLD appears in several places, including as a fallback for mapClusterMinZoom. I'm not sure they would stay in sync if a user changes their setting, but the environment variable stays at its default. Could you walk me through how you intended these to work together? My gut feeling is that it might be simpler to lean on the user setting alone and skip the environment variables, but I'm happy to be corrected if I'm missing something!

Also, while testing, I noticed some rough edges around the zoom level transitions. The point at which cluster groups break into individual markers and polylines take over didn't feel clean. I suspect the vs comparison in the unclustered-point filter might be involved. However, I'd love to hear your thoughts on the intended behavior before drawing any conclusions.

Regarding the two-query approach: I think there might be cases where the second query is sent unnecessarily, specifically at low zoom levels when includePolyline is false but diagonalFilter is set. This means that polylines are fetched but never rendered. At high zoom, where the diagonal filter is empty, both queries seem to hit the same set of trails. This seems like it could be collapsed into one query. Am I reading that correctly, or is there a reason to keep them separate?

palhaland:
Hi @slothful-vassal, than you for the review.

The reason in the first place for the two queries was for maintaining the cluster count, but you are right that that is wasting the bandwidth transferring them twice. I'm trying to modify the changes slightly now to transfer the id of the tracks to the client, and the client can cache the tracks so that it does not need to fetch them again. (and also keep the queries to a minimum towards the server). 

I also noticed when testing the discrepancy that you mentioned when zooming in and out, that the tracks should have been clustered, so I am trying to solve that as well. And it might be the cause of the confusing with the configuration parameters. I'll also update the pull request with a description of the changes.

palhaland:
@slothful-vassal, I'm having some issues understanding how to create clusters of the hidden tracks, right now that is missing, you only get the numbers when zooming all the way out (past the default 12 limit, or the user defined limit). If you have an idea, I don't mind a hand.

slothful-vassal (commented):
Thanks again for all your work on this! The concept feels coherent now, and I’ll gladly take a closer look in detail and test it thoroughly.

I do have one small but important concern: at the moment, we’re generally trying to move larger chunks of logic out of the client and into the server, so they can be reused for possible future extensions. I know you’ve already invested a lot of time here, and this would be another non-trivial refactor. But would you be interested in, and have the capacity for, moving the clustering logic into the server code?

That said, this is not a hard requirement for this PR. If it’s too much scope right now, we can also follow up separately later. My main thought is that doing it now could save us a fair amount of duplicate testing effort.

slothful-vassal (commented):
Will be implemented by #991.

PR #973, opened 2026-05-09
fix email change issues

slothful-vassal:
Fixes an inconsistent email-change flow where the web UI showed an error even though the email was already changed. The user stayed logged in with stale auth state, which then caused follow-up errors in later actions.

Email changes now use a dedicated `POST /api/v1/user/{id}/email` flow with `currentPassword`, refreshed auth state, `verified=false`, and a new verification email. Normal user updates no longer process email changes: the API rejects `email` on `POST /api/v1/user/{id}`, while the client strips `email` from regular `users_update` payloads to avoid accidentally sending full auth records.

### DIFF
--- a/db/hooks/users.go
+++ b/db/hooks/users.go
@@ -87,17 +87,6 @@ func UpdateUserHandler(client meilisearch.ServiceManager) func(e *core.RecordEve
 	}
 }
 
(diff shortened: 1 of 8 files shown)


Flomp (approved):
Fixed a small issue:
The email modal was not updating its value instantly. Meaning that if you clicked "Change email" again, you would still see the old value.

PR #980, opened 2026-05-10
initial commit

Flomp:
The changes to remote trail/list/comment fetching currently allow everyone with the link to access any trail/list/comment. This PR fixes that by checking the view/list rules before returning the result from pocketbase.

### DIFF
--- a/db/routes/remote_list.go
+++ b/db/routes/remote_list.go
@@ -62,6 +62,17 @@ func RemoteListGet(e *core.RequestEvent) error {
 		}
 	}
 
+	reqInfo, err := e.RequestInfo()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


slothful-vassal (commented):
RemoteTrailCommentsList: totalItems is still computed without applying the access rule, while items are filtered afterwards. So if any comments are filtered out, totalItems/totalPages may leak the number of comments on an inaccessible trail.

I consider this minor for this PR and approved it anyway. Your call whether to fix it now or in a follow-up.

Minor finding, not directly related to this PR: the comments ListRule/ViewRule may have a bug: `author = @request.auth.id` compares the comment author actor id with the authenticated user id. Should this be `author.user = @request.auth.id`, consistent with the update/delete rules?

slothful-vassal (approved):

PR #987, opened 2026-05-11
Fix remote trail sync for tagged trails

Flomp:
Currently, syncing remote trails fails for trails that have tags. E.g.: https://demo.wanderer.to/trail/view/@flomp@trails.tchncs.de/2e803f70a7632f7

This PR ensures that tag IDs are properly translated into local IDs. Tags that do not exist yet locally are automatically created.

### DIFF
--- a/db/routes/remote_trail.go
+++ b/db/routes/remote_trail.go
@@ -190,15 +190,65 @@ func syncTrailMetadata(app core.App, record *core.Record, data map[string]any) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


slothful-vassal (commented):
I found one small thing: if the tags are removed from the remote trail, that change is not synced locally. I think this is an edge case though, so I’ll approve this PR anyway.

slothful-vassal (approved):

PR #991, opened 2026-05-13
feat: implement server-side map clustering

palhaland:
## Description
This PR transitions map clustering from a client-side implementation to a server-side approach using Meilisearch for efficient querying and Supercluster on the SvelteKit server for geometry processing. This significantly improves performance for users with large trail datasets.

### Key Changes
- **Server-Side Clustering**: Added a new SvelteKit API route at `/api/v1/search/trails/cluster` that uses the `supercluster` library to process Meilisearch hits before they reach the browser.
- **Svelte 5 Reactivity Fixes**: Resolved bugs in `MapWithElevationMaplibre.svelte` where map initialization and data updates were not correctly triggered due to missing reactive state (`$state`) and incorrect destructuring of derived values.
- **MapLibre Validation Fixes**: Corrected issues where `minzoom`, `maxzoom`, and `filter` were being set to `undefined`, causing MapLibre rendering failures.
- **Dynamic Layer Visibility**: Removed hardcoded zoom constraints on map layers, allowing the backend-provided `bounding_box_diagonal` and frontend `tiers` filter to dynamically manage visibility for better performance.
- **Performance Optimizations**: Added basic caching for bounding box and filter values in the map route loader to reduce redundant API calls.
- **Font Correction**: Updated the cluster layer to use `Noto Sans Regular`, which is correctly hosted by the OpenFreeMap tileserver.


palhaland:
this one includes the bounding box feature as well and #963 can be closed if this one is accepted.

slothful-vassal (commented):
Thanks for working on this, server-side clustering is definitely the right direction here.

I still think we should avoid static zoom/diagonal cutoffs for deciding polyline visibility. The current approach makes the behavior depend on preconfigured thresholds, but the actual problem is result density: some viewports may contain only a handful of trails, while others contain hundreds at the same zoom level.

A dynamic threshold would be more robust: cluster all trail anchor points, then show polylines only for the top N currently visible trails by `bounding_box_diagonal`. N could default to something like 50-100 and maybe be configurable via env var or user setting. The effective diagonal threshold would then be derived from the current result set instead of configured manually.

This would also keep marker clustering conceptually separate from polyline visibility: all markers are clustered, and only the most relevant visible trails get route previews.

palhaland:
That was a pretty good idea, and this will allow all the tracks visible in
the viewport if it is less than the threshold. It also clean up the logic a
bit and made it a bit faster.

docker/docker-compose.dev.yml
```diff
@@ -36,6 +36,7 @@ services:
       start_period: 20s
       timeout: 10s
   web:
+    build: ../web
```
slothful-vassal: I’d prefer to keep only the env var change here unless we intentionally want to change the compose workflow.

web/src/lib/components/base/text_field.svelte
```diff
@@ -11,7 +11,7 @@
         error?: string | string[] | null;
         icon?: string;
         extraClasses?: string;
-        type?: "text" | "password" | "search";
+        type?: "text" | "password" | "search" | "number";
```
slothful-vassal: Is the TextField type="number" change still needed? I don’t see any usage of <TextField type="number"> in this PR, so it looks unrelated and can probably be dropped.

web/src/lib/components/trail/map_with_elevation_maplibre.svelte
```diff
@@ -808,6 +863,14 @@
             ...mapOptions,
         };
         map = new M.Map(finalMapOptions);
+        (window as any).map = map;
+
+        // Ensure we always have glyphs for symbol layers (like cluster counts)
```
slothful-vassal: What issue is the map-level setGlyphs handler fixing? ClusterLayer already defines glyphs in its layer spec, and mutating the active map style on every styledata event feels a bit broad. If this is needed for custom styles without glyphs, maybe we should keep it scoped/documented, or handle it in the layer/style manager instead of the map component.

web/src/lib/components/trail/map_with_elevation_maplibre.svelte
```diff
@@ -1,5 +1,6 @@
 <script lang="ts">
     import { page } from "$app/state";
+    import { env } from '$env/dynamic/public';
```
slothful-vassal: ?

web/src/lib/stores/trail_store.ts
```diff
@@ -534,7 +628,7 @@ function buildFilterText(user: AuthRecord, filter: TrailFilter, includeGeo: bool
     }
 
     if (filter.author?.length) {
-        filterText += ` AND author = ${filter.author}`
+        filterText += ` AND author = '${filter.author}'`
```
slothful-vassal: The `user.actor` filter quoting changes look reasonable, but they seem unrelated to server-side clustering. Since this PR is already touching quite a few moving parts, I’d prefer to handle these filter-syntax cleanups in a separate PR, unless they are required for the clustering endpoint to work correctly.

web/src/lib/vendor/maplibre-layer-manager/caret-layer.ts
```diff
@@ -9,6 +9,7 @@ export class CaretLayer implements BaseLayer {
         this.spec = {
             version: 8,
             name: "direction-carets",
+            glyphs: "https://tiles.openfreemap.org/fonts/{fontstack}/{range}.pbf",
```
slothful-vassal: Is the glyphs entry in CaretLayer needed? This layer only uses `icon-image` and does not render text, so I don’t think glyphs are required here. It may be better to keep glyph handling scoped to layers that actually use text, like the cluster count layer.

web/src/routes/map/+page.svelte
```diff
@@ -1,6 +1,7 @@
 <script lang="ts">
     import { browser } from "$app/environment";
     import { goto } from "$app/navigation";
+    import { env } from '$env/dynamic/public';
```
slothful-vassal: ?

slothful-vassal (commented):
I made a few fixes and UI improvements. I hope this is okay and in line with what you had in mind. Please take a look when you have a chance — I’d be happy to hear your feedback.

docker/docker-compose.dev.yml
palhaland: reverted this

palhaland:
The changes that you had made was ok for me, forgot to reply here.

slothful-vassal (approved):

PR #1007, opened 2026-05-20
Improve trail planning: option to re-order route anchors

slothful-vassal:
This PR replaces the old one (#629).

Similar functionality, but simpler approach without new dependency.

### Buttons
- hide 'Upload new file' when in edit mode (instead of disabled)
- draw/edit button moved before upload button to keep UI stable when 'Upload new file' is hidden in edit mode
image
image

### Trail anchor list
- drag-drop list added with all trail anchors
- undo/redo support for anchor reorder (improved)
- keyboard support
- icon is drag point → should work on touch devices too
- hovering the anchor list item, show the cumulated distances and elevation gains/losses
- long anchor names (reversed geocoded locations) are cut and animated when hovered

Bildschirmfoto vom 2026-05-20 21-06-59

### Search places
- in route edit mode, selecting a search result shows a button to add it as a new route endpoint
image

### POI
- in route edit mode, hovering a POI shows a button to add it as a new route endpoint
image


slothful-vassal:
Last changes:
- Anchor numbering now starts at the second anchor, since the first anchor is the start point.
- Start/end icons are now used consistently on the map as well.
- Map markers are highlighted when hovering or focusing items in the anchor list.
- Cumulative anchor stats now use the same metrics logic as the trail totals.
- Anchor location labels omit the country when all anchors are in the same country.

slothful-vassal:
I slightly adjusted the trail create/edit workflow.

- On the edit page, the route is now editable immediately. Requiring an extra “edit route” click there felt redundant.
- Uploading a new file for an existing route is more of an edge case, so the prominent upload/draw choice is kept for trail creation only.
- Since this might be inconvenient when you only want to change a few metadata fields, I’ve added the option to edit various metadata directly on the trail view page.
- For existing trails, replacing the route is now exposed through a more visible reset/replace action in the route toolbar.
- Destructive delete/reset icons now turn red on hover for clearer visual feedback.

image

<p>


image

<p>

image

<p>

image

<p>

image

Flomp:
This looks very sexy. Thank you!

Adding a POI as a new route endpoint if no other anchor has been defined causes an error to be shown:
```
can't access property "lat", previousAnchor is undefined
```
The anchor is added properly nonetheless.

I have one more small nitpick: I actually needed to look into the code to understand how I can trigger a reorder of the anchors. I expected the whole list item to trigger the drag, not just the anchor icon. Maybe with a grip icon (https://fontawesome.com/icons/classic/solid/grip-vertical) that appears on hover to make it clear?

slothful-vassal (commented):
That's a great point—thank you so much. I've tweaked the layout a bit. I hope you like it.

Flomp (approved):

PR #1012, opened 2026-05-23
fix: reduce Meilisearch load, debounce federation sync

### LINKED ISSUE #1009: Wanderer unresponsive and flooding memory
### What happened?

After updating to 0.19.1, Wanderer frequently becomes completely unresponsive. I have trouble understanding the issue but I found two potential leads:

1. wanderer-db frequently floods memory, is high on io and high on cpu
Image

2. I notice a very high count of requests against meilisearch, here is a quick excerpt on trying to build a statistic on this: 

```
58863 2026-05-17T20:16 GET /tasks/<id>
3657 2026-05-17T20:16 PUT /indexes/trails/documents
3109 2026-05-17T20:16 GET /keys
4 2026-05-17T20:16 GET /health
3 2026-05-17T20:16 POST /indexes/trails/documents
9796 2026-05-17T20:17 GET /tasks/<id>
517 2026-05-17T20:17 PUT /indexes/trails/documents
512 2026-05-17T20:17 GET /keys
32 2026-05-17T20:17 POST /indexes/trails/documents
4 2026-05-17T20:17 GET /health
2 2026-05-17T20:17 POST /indexes/trails/search
23221 2026-05-17T20:18 GET /tasks/<id>
2344 2026-05-17T20:18 GET /keys
1739 2026-05-17T20:18 PUT /indexes/trails/documents
27 2026-05-17T20:18 POST /indexes/trails/documents
4 2026-05-17T20:18 GET /health
2 2026-05-17T20:18 POST /indexes/trails/search
59852 2026-05-17T20:19 GET /tasks/<id>
3444 2026-05-17T20:19 PUT /indexes/trails/documents
3177 2026-05-17T20:19 GET /keys
```

..what I also notice is quite an amount of POST inbox federation requests. Not sure if cleaning up the remote actors in the database would help with that specifically but it would feel more like a dirty workaround, not a fix.

### Version

v0.19.1

slothful-vassal:
Fixes #1009 and several related issues:

- Polyline: Moved out of the Meilisearch indexer into a DB field, computed once in the create/update hook. Eliminates O(n) GPX filesystem reads on startup reindex.
- Trail update latency: Removed WaitForTask from UpdateTrail (fire-and-forget like create). Cuts ~500ms from every update. Better behavior on Meilisearch failure too — the DB save already succeeded before the hook fires, so users were getting error responses for operations that had actually worked.
- Federation sync storm: A missing concurrency guard caused unlimited parallel performFullSync goroutines per trail/list under load — maybe the source of the 58k Meilisearch task-poll requests/min in #1009. Fixed with a sync.Map per resource type. Staleness threshold now configurable via POCKETBASE_FEDERATION_SYNC_INTERVAL (minutes, default 60).


db/routes/remote_list.go
```diff
@@ -51,8 +51,15 @@ func RemoteListGet(e *core.RequestEvent) error {
...
 			updatedAt := record.GetDateTime("updated").Time()
-			if time.Now().UTC().Sub(updatedAt) > 60*time.Minute {
-				go performFullListSync(e.App, ctx, e.Request.URL, record)
+
+			iri := record.GetString("iri")
+			if time.Now().UTC().Sub(updatedAt) > remoteSyncThreshold {
+				if _, alreadySyncing := listSyncing.LoadOrStore(iri, struct{}{}); !alreadySyncing {
+					go func() {
+						defer listSyncing.Delete(iri)
+						performFullListSync(e.App, ctx, e.Request.URL, record)
```
palhaland: I asked Anitmatter-cli for a review:

> ### 1. Request Context Cancellation in Background Goroutines (Critical)
  In both remote_trail.go and remote_list.go, the handler spins off background goroutines for syncing:

    go func() {
        defer trailSyncing.Delete(iri)
        performFullSync(e.App, ctx, e.Request.URL, record)
    }()

  The Issue:
  The  ctx  passed to  performFullSync  is derived from  e.Request.Context()  via network.go. As
  soon as the HTTP handler returns (which happens immediately after starting the goroutine), the HTTP
  request context is canceled by the server.
  Inside  performFullSync , the HTTP client uses this canceled context for outbound API calls, causing
  the background sync to immediately fail with a  context canceled  error.
  The Fix:
  Derive a detached context that preserves values (like rate limit actor metadata) but does not cancel
  when the parent HTTP request completes:

    bgCtx := context.WithValue(context.Background(), "actor", ctx.Value("actor"))
    go func() {
        defer trailSyncing.Delete(iri)
        performFullSync(e.App, bgCtx, urlCopy, record)
    }()
  ──────

db/routes/remote_trail.go
```diff
@@ -64,8 +82,15 @@ func RemoteTrailGet(e *core.RequestEvent) error {
...
 			updatedAt := record.GetDateTime("updated").Time()
-			if time.Now().UTC().Sub(updatedAt) > 60*time.Minute {
-				go performFullSync(e.App, ctx, e.Request.URL, record)
+
+			iri := record.GetString("iri")
+			if time.Now().UTC().Sub(updatedAt) > remoteSyncThreshold {
+				if _, alreadySyncing := trailSyncing.LoadOrStore(iri, struct{}{}); !alreadySyncing {
+					go func() {
+						defer trailSyncing.Delete(iri)
+						performFullSync(e.App, ctx, e.Request.URL, record)
```
palhaland: Related to : https://github.com/open-wanderer/wanderer/pull/1012/changes#r3295257963

palhaland: ### 2. Race Condition on  e.Request.URL  (Medium)

  Passing the raw pointer  e.Request.URL  to a background goroutine is dangerous because the server may
  recycle or modify the HTTP request structure once the handler returns.

  The Fix:
  Clone the URL structure (or pass only the raw query string) before spawning the goroutine:

    urlCopy := *e.Request.URL // Dereference to copy
    go func() {
        defer trailSyncing.Delete(iri)
        performFullSync(e.App, bgCtx, &urlCopy, record)
    }()
  ──────

db/routes/remote_list.go
palhaland: ### 2. Race Condition on  e.Request.URL  (Medium)

  Passing the raw pointer  e.Request.URL  to a background goroutine is dangerous because the server may
  recycle or modify the HTTP request structure once the handler returns.

  The Fix:
  Clone the URL structure (or pass only the raw query string) before spawning the goroutine:

    urlCopy := *e.Request.URL // Dereference to copy
    go func() {
        defer trailSyncing.Delete(iri)
        performFullSync(e.App, bgCtx, &urlCopy, record)
    }()
  ──────

db/main.go
```diff
@@ -213,10 +214,35 @@ func registerCronJobs(app core.App, client meilisearch.ServiceManager) {
...
 	initMeilisearchConfig(client)
-	go initMeilisearchDocuments(app, client)
+	go func() {
+		backfillPolylines(app)
+		initMeilisearchDocuments(app, client)
+	}()
 	return nil
 }
 
+func backfillPolylines(app core.App) {
```
palhaland, line 224: ### 2. Critical Issues Identified

  #### ⚠️ The Pagination Skipped-Record Bug in  backfillPolylines

  The current pagination mechanism in main.go uses offset-based pagination while modifying the
  records to no longer match the filter condition:
    err := app.RecordQuery("trails").
        AndWhere(dbx.NewExp("(polyline IS NULL OR polyline = '') AND gpx != ''")).
        Limit(pageSize).
        Offset(page * pageSize).
        All(&trails)
  Why this fails:

  1. On Page 0 ( page = 0 ), it retrieves the first 100 matching trails and successfully saves them
  with computed polylines.
  2. Since these 100 trails now have a non-empty  polyline  value, they are removed from the subset
  matching the query condition.
  3. On Page 1 ( page = 1 ), the query is executed with  Offset(100) . Because the first 100 matching
  records were removed, this skips the next 100 matching records (positions 101–200, which are now at
  indices 1–100 of the matching subset).
  4. As a result, approximately 50% of the matching trails are skipped during the migration backfill.
  #### Fix:

  Use keyset pagination (ordering by ID and filtering on  id > lastSeenID ). This guarantees all
  records are processed regardless of mutations:

    func backfillPolylines(app core.App) {
        const pageSize int64 = 100
        var lastID string
        for {
            q := app.RecordQuery("trails").
                AndWhere(dbx.NewExp("(polyline IS NULL OR polyline = '') AND gpx != ''")).
                OrderBy("id ASC").
                Limit(pageSize)

            if lastID != "" {
                q = q.AndWhere(dbx.NewExp("id > {:lastID}", dbx.Params{"lastID": lastID}))
            }

            trails := []*core.Record{}
            err := q.All(&trails)
            if err != nil || len(trails) == 0 {
                break
            }

            for _, r := range trails {
                if err := util.SavePolyline(app, r); err != nil {
                    log.Printf("backfill polyline failed for trail %s: %v", r.Id, err)
                }
                lastID = r.Id
            }
        }
    }
  ──────

palhaland (dismissed):
I got some help from Antigravity-CLI (Model: Gemini 3.5 Flash – Medium)

for some review comments

slothful-vassal (commented):
Thanks for the review and the clear explanation. All findings were spot on, and the suggested fixes made them easy to address.

Flomp (approved):
lgtm

PR #1014, opened 2026-05-23
fix: use ActivityPub actor ID for shared trails/lists search token filtering

palhaland:
Is it the incorrect id used for comparing with the shared user?


Flomp:
Thanks for the PR. You are correct. I removed the token creation in the CreateUserHandler. It was deprecated in v0.19.0.

Flomp (approved):

PR #1019, opened 2026-05-26
feat(web): replace vendored exif-js with maintained exifreader package

Vertux:
The vendored, unmaintained exif-js did not reliably read GPS tags from modern photos, so waypoints created "from photo" got no coordinates.

Drop web/src/lib/vendor/exif-js/ and use the maintained `exifreader` package. A single helper gpsFromPhoto(source: string | File) in web/src/lib/models/gpx/utils.ts reads coordinates defensively across several shapes (ExifReader's computed `gps` group, raw EXIF DMS rationals, and a non-expanded fallback), returning null when there are no usable GPS tags. Both call sites (waypoint modal and the trail editor) use it.


slothful-vassal:
Hi,

PR #1021, opened 2026-05-26
feat(db): resilient HTTP client + correctness fixes for integration s…

Vertux:
…yncs

Outbound HTTP from the nightly integration sync has been hardened without introducing new endpoints or user-facing controls.

Shared resilient HTTP client (db/util/httpclient.go):
- 60s timeout on each request.
- DoWithRetry retries on network errors, HTTP 429, and 5xx with Retry-After honoring + exponential backoff with jitter.
- ReadAllLimited caps response bodies at 100 MiB to protect against pathologically large responses (e.g. an unexpectedly huge GPX export).

All Strava/Komoot/Hammerhead read paths now go through the new client. Uploads (Hammerhead activity upload) stay on a timeout-only client and do NOT retry — retrying an upload risks creating duplicate activities.

Correctness fixes that were quietly swallowing errors:
- Strava: per_page=50 pagination, stopping when a partial page is returned (previously stopped only on an empty page, requiring an extra round-trip).
- Strava: per-user decrypt/marshal/save errors now log and `continue` to the next user instead of aborting the whole nightly run.
- Strava: invalid `After` date no longer aborts the run.
- Komoot/Hammerhead: previously unchecked json.Unmarshal errors are now logged or returned with context.

### DIFF
--- a/db/integrations/hammerhead/hammerhead.go
+++ b/db/integrations/hammerhead/hammerhead.go
@@ -92,11 +92,12 @@ func SyncHammerhead(app core.App, client meilisearch.ServiceManager) error {
 		if hammerheadIntegration.After != "" {
 			t, err := time.Parse("2006-01-02", hammerheadIntegration.After)
 			if err != nil {
-				return err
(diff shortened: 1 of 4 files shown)


slothful-vassal:
Hi,

PR #1022, opened 2026-05-26
feat(db): hardening — sanitizer + ActivityPub GET rate limit

Vertux:
Two small, defensive improvements:

HTML sanitizer (db/util/sanitize.go): restrict the allowed value of <a target> to "_blank" only. Previously any value passed through, so sanitized markup could re-target the wanderer window or arbitrary named windows.

Per-client-IP rate limiting on the public ActivityPub GET routes (db/util/network.go, wired in db/main.go): /activitypub/actor, /activitypub/actor/{id}/{follow}, /activitypub/trail/{id}, and /activitypub/comment/{id} now share a per-IP token bucket so scraping or flooding the public endpoints from a single host is throttled. The signature-verified POST inbox is intentionally left unthrottled so legitimate federation delivery bursts are not dropped.


slothful-vassal:
Hi,

PR #1023, opened 2026-05-26
fix(web): self-heal a stale Meilisearch tenant token in hooks.server.ts

Vertux:
After a Meilisearch reinstall or master-key rotation, previously issued tenant tokens cached in the `meilisearch_token` cookie become invalid and every search returns 403 until the cookie expires (~24h by default), with no visible error in the UI.

hooks.server.ts now probes the token with a cheap request against the configured Meilisearch host. If Meilisearch rejects it (401/403) the cookie is dropped and a fresh token is fetched on the next request. A successful probe, a network error, or a missing host all pass through untouched, so a working token is never discarded.

### DIFF
--- a/web/src/hooks.server.ts
+++ b/web/src/hooks.server.ts
@@ -50,6 +50,30 @@ function isFormContentType(request: Request) {
   );
 }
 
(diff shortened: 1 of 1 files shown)


slothful-vassal:
Hi,

PR #1024, opened 2026-05-26
feat(db): optional superuser bootstrap from environment variables

Vertux:
Allow provisioning the first PocketBase superuser at startup from POCKETBASE_SUPERUSER_EMAIL / POCKETBASE_SUPERUSER_PASSWORD, so a fresh container can come up without the operator having to follow the interactive installer link.

Opt-in: both variables must be set; neither set is a no-op. Setting only one is treated as a misconfiguration and logged.

Idempotent: the account is created only if no superuser with that email already exists. An existing password is never reset, so leaving the variables in the environment is safe.

Defensive: any failure (collection lookup, save) is logged via app.Logger().Error() and does not abort startup, so a misconfigured bootstrap can never prevent the app from coming up.

### DIFF
--- a/db/hooks/bootstrap.go
+++ b/db/hooks/bootstrap.go
@@ -42,6 +42,59 @@ func OnBootstrapHandler() func(se *core.BootstrapEvent) error {
 			e.App.Settings().SMTP.Password = v
 		}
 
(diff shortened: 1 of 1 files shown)


slothful-vassal:
Hi,

Issue #1026, opened 2026-05-29
OSM OIDC login: default scopes rejected, and display name discarded


mfuhrmann:
### What happened?

When configuring wanderer with OpenStreetMap (OSM) as an OIDC provider, newly created users receive a numeric ID as their username instead of their OSM display name.

Without a scope override, OSM login fails outright; with one, it succeeds but the username is discarded.

Root causes:

1. PocketBase's default OIDC scopes include email and profile, which OSM rejects. wanderer needs a way to override these scopes per deployment.
2. Even with correct scopes (`openid read_prefs`), OSM returns `preferred_username` but PocketBase's internal `oldCanAssignUsername` check silently fails, causing username assignment to be skipped and a numeric fallback to be used.

### Version

`v0.19.1`, `v0.20.0`, `dev`

### Why you should add OSM auth

Your tool is based on OSM data, and a lot of the auth providers you offer [here](https://wanderer.to/use/authentication/#oauth2) are typical big tech players that, especially OSM people do not use by default 😂
Even new users could get a recommendation to create an OSM account. By going this way, the chance is higher to get people hooked into the OSM world. A lot of services can use the OSM account.

PR #1027, opened 2026-05-29
feat(db): allow overriding OIDC scopes via OIDC_SCOPES

mfuhrmann:
## What changed

Adds an `OIDC_SCOPES` environment variable for the `db` service. When set, it replaces the scopes requested by the `oidc`, `oidc2` and `oidc3` providers with a comma-separated list. When unset, the PocketBase defaults are used, so existing installations are unaffected.

## Why

wanderer requests `openid`, `profile` and `email` from every OIDC provider. Some providers reject an authorization request that carries scopes they do not recognise, rather than ignoring the unknown ones — the login then fails before the user ever reaches the consent screen.

OpenStreetMap is such a provider: it accepts neither `profile` nor `email`, which makes it unusable as an OIDC provider today. With `OIDC_SCOPES: "openid,read_prefs"` it works.

PocketBase itself offers no way to configure this. `core.OAuth2ProviderConfig` (v0.38.0) exposes `PKCE`, the endpoint URLs, client credentials, display name and an `Extra` map, but no scope field — scopes come from the provider factory in `tools/auth`, which is why this is an override of the factory rather than a settings change.

## How it was tested

- `db/main_test.go` covers the unset case, a single scope, a comma-separated list, surrounding whitespace, empty entries, and a separators-only value. The last two fall back to the provider defaults.
- Running against a live OpenStreetMap OIDC app on a self-hosted instance: login fails without the variable and succeeds with `openid,read_prefs`.

## Notes

Whitespace around entries is trimmed and empty entries are dropped, so `"openid, read_prefs"` behaves as expected.

The scope list applies to all three OIDC provider slots. Configuring them separately would mean three variables; that seemed like the wrong trade until someone actually needs two different OIDC providers with two different non-default scope sets.

This PR previously also mapped the OAuth2 username from `preferred_username`. That has been split out — testing showed it needs sanitising against the `username` field pattern before it is safe, since OSM display names may contain spaces and other characters the field rejects. It will follow as a separate PR.

Addresses root cause 1 of #1026. Root cause 2 (the discarded display name) is left for a separate PR.

### DIFF
--- a/db/main.go
+++ b/db/main.go
@@ -5,12 +5,14 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 
 	"github.com/meilisearch/meilisearch-go"
 	"github.com/pocketbase/dbx"
 	"github.com/pocketbase/pocketbase"
 	"github.com/pocketbase/pocketbase/core"
 	"github.com/pocketbase/pocketbase/plugins/migratecmd"
+	"github.com/pocketbase/pocketbase/tools/auth"
 
 	"pocketbase/commands"
 	"pocketbase/hooks"
@@ -67,6 +69,8 @@ func main() {
 
 	setupCommands(app)
 
+	configureOIDCScopes()
+
 	if err := app.Start(); err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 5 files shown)


mfuhrmann:
@Flomp I rebased #1027 onto `dev`. The conflict is resolved and this PR now covers only the `OIDC_SCOPES` change. The username half is split into #1159. Both address #1026.

slothful-vassal (changes requested):
Thanks for the PR and the clear explanation! The implementation looks straightforward.

My main concern is applying the override to all three OIDC providers. This already affects an installation using **OSM alongside another provider with default scopes**. Setting `openid,read_prefs` for OSM would also change the second provider’s request, potentially causing rejected scopes or missing email/profile information.

I understand the motivation to keep configuration simple, but could we make the scopes configurable per provider slot, keeping PocketBase’s defaults for any slot without an override? I think that would make this useful without disrupting another configured login provider.

Two smaller documentation suggestions:

- According to the [[OSM documentation](https://wiki.openstreetmap.org/wiki/OAuth#Using_OpenStreetMap_as_identity_provider)](https://wiki.openstreetmap.org/wiki/OAuth#Using_OpenStreetMap_as_identity_provider), `openid` alone is sufficient for OIDC login. Could we use that in the example? `read_prefs` grants additional access to user preferences that this flow doesn’t appear to need.
- The generated-username note seems a little too broad. PocketBase already maps `preferred_username` when the configured mapping exists and the value passes validation and uniqueness checks. It would be more accurate to say that OSM accounts **may** receive a generated username when their display name cannot be used.

mfuhrmann:
Thanks for the review! I made the changes:

- One setting per provider: `OIDC_SCOPES` now only changes `oidc`, `OIDC2_SCOPES` only `oidc2`, and `OIDC3_SCOPES` only `oidc3`. A provider without a setting keeps the PocketBase defaults, so OSM no longer affects the other providers
- OSM example: the docs now use `openid` only
- Username note: now says OSM accounts may get a generated username

I also updated the test to check each provider on its own.

slothful-vassal (approved):

PR #1044, opened 2026-06-04
fix self-federation

### LINKED ISSUE #1045: NeedsFullSync can be true for local trails causing 500
> I have the same issue. I just created a trail on the demo instance. Put it on "public". If it is private it still works. And after putting it on "public" I get this Error 500 with these tokens. The same on the demo instance: https://demo.wanderer.to/trail/view/@demo/5ea09e879a685a0 or this https://demo.wanderer.to/trail/view/@demo/8fc4224e1bb5fbc 

 _Originally posted by @docnero in [#1043](https://github.com/open-wanderer/wanderer/issues/1043#issuecomment-4619319047)_

This is a separate issue from #1043, I think. Can you please provide steps to reproduce? I can see the result on the demo page; however, I'm struggling to recreate the issue.

slothful-vassal:
Viewing some public trails returned `500 "Sync failed."` because the instance tried to federate its **own** content from itself: an inbound activity (announce/comment) on a local trail flags it `needs_full_sync`, and the next view runs `performFullSync` against our own origin (loopback → SSRF 500, and `needs_full_sync` never clears → permanent 500).

Fix (guard via new `util.IsLocalIRI`):
- **Read** (`performFullSync`/`performFullListSync`): empty or own-origin IRI → skip fetch, clear `needs_full_sync`, return local record (self-heals poisoned records); a missing local record now returns 404 instead of an unsaved shell.
- **Ingest** (`TrailFromActivity`/`ListFromActivity`): local IRI only resolved for a local actor; a remote actor referencing a local object id is rejected (prevents feed/share/notification spoofing on local content).

**Affected versions:** the bug exists since **v0.19.0** (federation refactor #930). It is not new in 0.19.2 — but 0.19.2 made background re-syncs actually run (`context.Background()` instead of the cancelled request context), which increases how often it's triggered.

### DIFF
--- a/db/routes/remote_list.go
+++ b/db/routes/remote_list.go
@@ -49,6 +49,12 @@ func RemoteListGet(e *core.RequestEvent) error {
 				}
 				return e.InternalServerError("Sync failed", err)
 			}
+			if record.Id == "" {
+				// Local content that does not exist: performFullListSync
+				// short-circuits local IRIs and returns the unsaved shell —
+				// surface a real 404 instead of access/expand on a missing record.
+				return e.NotFoundError("List not found", nil)
+			}
 		} else {
 			updatedAt := record.GetDateTime("updated").Time()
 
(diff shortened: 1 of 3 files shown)


Flomp (approved):

Guacam-Ole:
Thanks for fixing this. That drove me crazy. Checked the db and had about 3 Million Duplicates :)

PR #1046, opened 2026-06-04
Add IRI to local resources

Flomp:
This PR adds IRIs to all local resources. Existing resources are accounted for via a migration.

How to test:
- [x] Validate that all existing comments/lists/summit_logs/trails/waypoints have a valid IRI after startup.
- [x] Validate that all new comments/lists/summit_logs/trails/waypoints receive a valid IRI upon creation.
- [x] Validate that the following federated actions still work:

1. Trail/List Share
2. Summit Log Create
3. Comment Create
4. Trail Like/Unlike
5. Remote trail fetch
6. Remote comment fetch
7. Remote list fetch


db/hooks/trails.go
```diff
@@ -26,18 +28,31 @@ func CreateTrailHandler(client meilisearch.ServiceManager) func(e *core.RecordEv
...
 		}
 
+		// add local iri
+		origin := os.Getenv("ORIGIN")
+		if origin == "" {
+			return fmt.Errorf("ORIGIN not set")
+		}
+		if e.Record.GetString("iri") == "" {
+			e.Record.Set("iri", fmt.Sprintf("%s/api/v1/trail/%s", origin, e.Record.Id))
+			if err = e.App.Save(e.Record); err != nil {
```
slothful-vassal: This triggers the trail update hook, resulting in an update federation before the create federation has finished.

edit: same for list.go

db/migrations/1780566579_add_iri_to_local_resources.go
```diff
@@ -0,0 +1,154 @@
...
+			if err != nil {
+				return err
+			}
+			authorIri, _ := url.Parse(author.GetString("iri"))
+			authorDomain := author.GetString("domain")
+			authorScheme := authorIri.Scheme
+			iri = fmt.Sprintf("%s://%s/api/v1/list/%s", authorScheme, authorDomain, l.Id)
+			l.Set("iri", iri)
+
+			if err := app.Save(l); err != nil {
```
slothful-vassal: app.Save will trigger the list update hook here, causing federation spam for existing lists during migration

db/migrations/1780566579_add_iri_to_local_resources.go
```diff
@@ -0,0 +1,154 @@
...
+			iri := t.GetString("iri")
+			if iri != "" {
+				continue
+			}
+			author, err := app.FindRecordById("activitypub_actors", t.GetString("author"))
+			if err != nil {
+				return err
+			}
+			authorIri, _ := url.Parse(author.GetString("iri"))
+			authorDomain := author.GetString("domain")
```
slothful-vassal: This migration derives the IRI from the actor domain/scheme instead of ORIGIN, so existing local trails can get a different base URL than newly created trails. For example with ORIGIN=http://localhost:5173, migrated trails become http://localhost/api/v1/trail/... while new trails become http://localhost:5173/api/v1/trail/....

db/util/activitypub.go
slothful-vassal: Shouldn't we use the persisted trail iri here instead of rebuilding the object ID again? This would avoid potential mismatch issues.

slothful-vassal (changes requested):

db/util/activitypub.go
Flomp: Addressed in [fc7cfed](https://github.com/open-wanderer/wanderer/pull/1046/commits/fc7cfede4ce70c6883f9d83fb0f7ed22fa948b22)

db/hooks/trails.go
Flomp: Ah tricky. So UnsafeWithoutHooks?

db/migrations/1780566579_add_iri_to_local_resources.go
Flomp: Probably here, UnsafeWithoutHooks as well then.

db/hooks/trails.go
slothful-vassal: yes, unsafe would work, but maybe it's more elegant to set the iri before the save?

untested (the id should be set already):

app.OnRecordCreate("trails").BindFunc(func(e *core.RecordEvent) error {
    if e.Record.GetString("iri") == "" {
        e.Record.Set("iri", fmt.Sprintf("%s/api/v1/trail/%s", origin, e.Record.Id))
    }

    return e.Next()
})

db/migrations/1780566579_add_iri_to_local_resources.go
slothful-vassal: yes, here you need unsafe

db/hooks/trails.go
Flomp: Unfortunately, the id is only set after. e.Next() is called.

That's why I pushed it below.

db/migrations/1780566579_add_iri_to_local_resources.go
Flomp: Fixed in [f79e4dd](https://github.com/open-wanderer/wanderer/pull/1046/commits/f79e4dd6cf68e4bb4e3cd6a146dda8433782fbb6)

db/migrations/1780566579_add_iri_to_local_resources.go
Flomp: I think it's an edge case that actor iri != origin. Nonetheless fixed in [f79e4dd](https://github.com/open-wanderer/wanderer/pull/1046/commits/f79e4dd6cf68e4bb4e3cd6a146dda8433782fbb6)

Flomp:
Since the `id` is not set before `e.Next()` is called, we now use `UnsafeWithoutHooks()` to prevent the Update hook from being called when saving the `id` in the Create hook.

slothful-vassal (dismissed):

Flomp:
> found an issue in federated actions tests - details will follow on discord

This is fixed now. I was too eager in `findLocalList/TrailByRemoteInfo` and only looked up the IRI to find already existing trails/lists. But since the ID between the remote instance and the local one differs, it failed to find local copies of the trail/list.

slothful-vassal:
1. Create a new trail list and save
2. edit the trail list and add some trails and save the changed list
3. open the trail list page

=> the list card shows no statistics

image

Flomp:
Sorry this is so sloppy...

I fixed two places in the last commit that relied on no IRI = local. Now we do an additional check if the associated actor is indeed local.

Since this still uses the old "isLocal" syntax, it would be good to merge this PR before #1048.

slothful-vassal (approved):

PR #1048, opened 2026-06-04
Feat add actors to meilisearch

Flomp:
## Actor search via Meilisearch

Replaces the PocketBase `SELECT` query used for actor search with a dedicated Meilisearch `actors` index. For large deployments with many federated actors, the previous approach could become noticeably slow — this change brings actor search in line with the rest of the search infrastructure.

### What changed

- All `activitypub_actors` records are now indexed in a new Meilisearch index called `actors`.
- The `/api/v1/search/actor` endpoint queries this index instead of PocketBase directly.

### Non-breaking

This is a non-breaking change. Existing behaviour is preserved; the index is additive.

> ⚠️ **Heads up for large instances:** If your `activitypub_actors` table is large (e.g. a heavily federated instance), the initial index build may consume meaningful disk space and take some time to complete. Plan accordingly before deploying.


slothful-vassal:
When searching for an actor in trails list:

{
    "message": {
        "name": "MeilisearchApiError",
        "cause": {
            "message": "The provided tenant token cannot acces the index `actors`, allowed indexes are [\"lists\", \"trails\"].",
            "code": "invalid_api_key",
            "type": "auth",
            "link": "https://docs.meilisearch.com/errors#invalid_api_key"
        },
        "response": {}
    }
}

slothful-vassal (changes requested):
fix unauthorized tenant token issue

Flomp:
1. Added search token versioning to make PR non-breaking.
2. Made `id` filterable.

web/src/lib/components/base/editor.svelte
slothful-vassal: isLocal vs is_local? I see '@user@instance' instead of '@user' for local user mentions in comments.

slothful-vassal (changes requested):
There are further code paths where `isLocal` is still used instead of `is_local`. This breaks the share flow for local actors.

Flomp:
Thanks for the review!
Two fixes:
1. pocketbase variables should be snake_case anyway, so I chose to rename the "isLocal" column to "is_local" and change all occurrences that are not in past migrations.
2. Actor IRIs are now indexed, which fixes the issue that sharing did not work

slothful-vassal (changes requested):
1. If an anonym user is searching for an actor on the homepage, I see a 401 error in browser dev console (minor issue)


2. Federated instance required:
- open public trail
- open share modal
- search for remote actor, eg. '@demo@demo.wanderer.to'
- select the remote actor
-> does not work, in browser dev console:

Flomp:
1. That is the intended behaviour. I would like to prevent bots from indexing the actor table.
2. Can you please double-check that you are on the latest version? This is the behaviour I changed with [2800c37](https://github.com/open-wanderer/wanderer/pull/1048/commits/2800c37f7d0894e76dab31c2161ebf5873437ee5). I cannot replicate the error on my end in the latest version. Can you also check, please, whether your documents in the actor index in meili have an iri field?

slothful-vassal:
1. Sorry. I didn't mean to cast doubt on the concept. I was just wondering if an (hidden) error message should be generated or if the actor search should be blocked on the client side. But, as I said, it's not important and won't prevent the merge.
2. Yes, I checked again, and I'm still getting the reported errors when following the steps I mentioned. What am I doing wrong?
Edit: Perhaps “does not work” wasn't precise enough: the actor is found and can be selected, but it isn't applied.

Flomp:
Could you please check whether the documents in your `actors` index have an `iri` field?

Flomp:
Thanks for clarifying on discord. This is fixed now.

slothful-vassal (approved):

PR #1052, opened 2026-06-07
fix: prevent duplicate feed entries on repeated trail/list activities

Guacam-Ole:
## Problem

`InsertIntoFeed` (db/util/feed.go) inserts a new `feed` row on every call without checking whether the `(actor, item)` pair already exists.

On current `main` this is reachable through the inbox: `processCreateOrUpdateTrailActivity` and `processCreateOrUpdateListActivity` call `InsertIntoFeed` unconditionally, so **every inbound `Update` activity for an already-known remote trail/list appends another copy of the same feed entry** for each local follower. Long-running instances following active remote authors accumulate duplicates steadily; the feed endpoint then has to page over them on every home-page load.

This is a followup to the fixed  #1044 fixed issue. The fix works, but we still have a single duplicate each time. (#1044 fixes the main issue of an infinite loop. this PR just takes care of the rest)

## Fix

If a feed entry for the same `(actor, item)` pair already exists, return the existing record instead of inserting a duplicate. One file changed, no behaviour change for first-time inserts.


Flomp:
Thanks for the fix. We made two follow-up changes on top of your PR:

In `db/util/feed.go`: The original guard `(if err == nil && existing != nil)` silently swallowed genuine database errors. If `FindFirstRecordByFilter` failed for any reason other than "no rows found," the error was discarded and the function fell through to attempt an insert anyway. We tightened this to explicitly check for `sql.ErrNoRows`: a not-found error means no duplicate exists and we proceed to insert; any other error is propagated to the caller rather than ignored.

In `db/migrations/1782901091_updated_feed.go` (new): The application-level check alone has a race condition. Two concurrent federation activities for the same `(actor, item)` can both pass the guard before either inserts. We added a unique index on `(actor, item)` so the database enforces the constraint even if the race fires. Because SQLite refuses to build a unique index when duplicates already exist, the migration first deletes all but the latest row per pair (using `MAX(rowid)`, since SQLite rowids are monotonically increasing) before creating the index. Instances that accumulated duplicates before this fix will be cleaned up automatically on next startup.

Flomp (approved):

PR #1059, opened 2026-06-13
feat: Refine trail category model

### LINKED ISSUE #439: Bug in mapping Komoot categories

### LINKED ISSUE #591: Custom trail categories not internationlaized in user frontend (like default categories)
### What happened?

My instance is used (mainly) in German.
I added some custom trail-categories via the admin-interface and noticed all default categories are in English but displayed in German (eg Hiking -> Wandern) in the frontend. So i added my own categories in English, but they are still displayed in their original English name in the frontend (for Rennradfahren i tested Roadbiking and Cycling).

(Not a serious bug as i may use German category names.)

Image

### Version

v 3 (acc. to docker-compose)

slothful-vassal:
## Motivation

wanderer ships with a small, fixed set of trail categories. That keeps the model simple, but leaves no good way to capture more specific activity types: you either overload a broad category or add a custom one — and custom categories don't federate reliably across instances.

Part of the friction is that a category's `name` does double duty. It's the matching key for integrations and federation, and at the same time the frontend uses it as an i18n key for display. That's fine for the built-in categories, but awkward for custom ones: the name has to stay stable so matching keeps working, yet display labels need somewhere of their own to live and be translated.

And because new trails are created without a category today, if the user had not set his "favourite sport", they often stay uncategorized; a per-user priority order lets the most relevant category be pre-selected from the start.

## Summary

- Keeps the category `name` as the canonical matching identity, now matched on a normalized form (case/accent/separator-insensitive).
- Adds optional, parent-scoped subcategories for finer-grained activity types.
- Moves category and subcategory display to instance-managed `translations`, falling back to English and then to the canonical name.
- Matches federated categories and subcategories by normalized name across both ActivityPub and direct remote sync, degrading to the parent when a subcategory is unknown; raw remote values are persisted for later backfill.
- Adds per-user category preferences — hide from search, and trail creation/editing, exclude federated trails, and priority ordering. Priority also drives the category pre-selected on new trails and the profile's "favourite sport".
- Replaces the former standalone "favourite sport" setting with priority: the profile now shows the highest-priority category read-only with a link to manage it, and existing favourites are migrated to priority 1.
- Indexes category/subcategory IDs (plus `is_federated` and raw remote values) in Meilisearch and applies per-user preference filters server-side across all query paths.
- Adds the supporting UI: a category/subcategory picker in the trail form, an icon-based filter with a subcategory overlay (hover / long-press) including a "no subcategory" option, a bulk-edit modal for category/subcategory/difficulty, and a Settings → Categories page for visibility, federated-trail handling, ordering, and per-subcategory preferences.
- Handles migration and backfill: de-duplicates category names that collide after normalization (keeping the oldest, renaming the rest), persists raw remote values, migrates favourite sport → priority, and drops the unused category image field.

### Settings → Categories
image

### Bulk-edit modal

image

### Trail filter

image

## Test Points

**Trail form**
- [x] Create/edit a trail: pick a broad category, then a subcategory; confirm it saves and the card/list/detail show the icon, badge, and label.
- [x] A category/subcategory without a translation in the active UI language falls back to English, then to the raw name.

**Filtering**
- [x] Filter by a category; open the subcategory overlay (hover on desktop, long-press on touch); filter by a specific subcategory and by “no subcategory”; confirm results match.

**Bulk edit**
- [x] Select multiple trails → Adjust → set category/subcategory/difficulty; confirm all selected trails update.

**Settings → Categories**
- [x] Toggle **Show** off: category disappears from search/discovery and from the trail-form picker (and its federated trails are hidden).
- [ ] Toggle **Remote trails** off: federated trails in that category disappear, your own trails in it remain.
- [x] Reorder categories and hide a subcategory; confirm the order/hiding reflects in the picker and filter.
- [x] Preferences are per-user: a second account is unaffected.

**Search visibility**
- [ ] Logged-in user with exclusions vs. anonymous request (no filtering); local vs. federated trails behave per preferences.

**Federation & imports**
- [ ] Sync/federate a trail with a category + subcategory → matches by normalized name; unknown subcategory degrades to parent only; unknown category leaves the trail uncategorized but visible; `remote_category`/`remote_subcategory` are persisted.
- [ ] After an admin adds the matching (sub)category, the backfill assigns it to previously stored trails.
- [ ] Import from Strava/Komoot (e.g. a run) maps to the expected category (Running).

**Deploy**
- [ ] Fresh instance seeds defaults (incl. Running + default subcategories); existing instance gains Running; Meilisearch reindex picks up `category_id`/`subcategory_id`/`is_federated`.


Flomp:
Thank you! Looks great as usual overall. Findings below.

**Status key:**  ✅ Pass | ❌ Fail | ⚠️ Partial

---

## A. Trail Form — Category Picker

| # | Test | Status | Notes |
|---|------|--------|-------|
| A1 | Create a new trail — category picker appears and shows all enabled categories with icons | ✅ | |
| A2 | Select a broad category, then pick a subcategory from the secondary picker | ✅ | |
| A3 | Save the trail — trail card, list view, and detail view all show the correct icon, badge, and label | ✅ | |
| A4 | Edit an existing trail — previously set category/subcategory is pre-populated | ✅ | |
| A5 | Switch language to German — category and subcategory labels appear in German; switch to a language without a translation — falls back to English, then to the raw name | ✅ | |
| A6 | New trail is created without manually picking a category — highest-priority category from the user's preferences is pre-selected automatically | ❌ | `/web/src/routes/trail/edit/[id]/+page.ts:32` passes `categories[0]` to `new Trail()`, so `/web/src/routes/trail/edit/[id]/+page.svelte:196–197` returns early and the preference-ordered logic at line 200–207 is never reached. |

---

## B. Filtering

| # | Test | Status | Notes |
|---|------|--------|-------|
| B1 | On the map/search page, click a category icon to filter — results show only that category | ✅ | |
| B2 | Hover (desktop) or long-press (touch) a category filter — subcategory overlay appears | ✅ | |
| B3 | Select a specific subcategory in the overlay — results narrow to that subcategory | ✅ | |
| B4 | Select "no subcategory" option in the overlay — results show only trails with no subcategory set | ✅ | |
| B5 | Click an active category filter again to deselect — all categories are shown again | ✅ | |

---

## C. Bulk Edit

| # | Test | Status | Notes |
|---|------|--------|-------|
| C1 | Select multiple trails via the list → open Adjust/bulk-edit modal — category, subcategory, and difficulty fields are present | ✅ | |
| C2 | Set a new category + subcategory in the modal and confirm — all selected trails are updated | ✅ | |
| C3 | Change difficulty only (leave category unchanged) — only difficulty updates, category is untouched | ✅ | |

---

## D. Settings → Categories

| # | Test | Status | Notes |
|---|------|--------|-------|
| D1 | Navigate to Settings → Categories — page loads with all categories listed | ✅ | |
| D2 | Toggle **Show** off for a category → confirm modal appears → confirm — category disappears from trail-form picker and from search/filter bar | ✅ | |
| D3 | Toggle **Show** off — trails in that category are hidden from search and discovery | ✅ | |
| D4 | Toggle **Remote trails** off for a category — federated trails in that category disappear; your own trails in it remain visible | ✅ | |
| D5 | Reorder categories via drag-and-drop — new order is reflected in the trail-form picker and filter bar | ✅ | |
| D6 | Hide a specific subcategory in the preferences — subcategory no longer appears in picker or filter | ✅ | |
| D7 | Log into a second user account — verify it has independent preferences (changes from D2–D6 do not affect the second user) | ✅ | |

### D — Bugs Found

- **Stale plugin warning on category disable:** Users see a warning about plugin mappings even when no plugins are installed. `shouldConsiderPluginInstance` in `settings/categories/+page.svelte:133–134` considers an instance relevant if `enabled || hasAuthInfo`. A previously-installed-then-deleted plugin whose DB record survives with stale auth credentials satisfies `hasAuthInfo`, so it still triggers the inactive-plugin warning branch. Fix: wipe auth fields on plugin deletion so `hasAuthInfo` returns false for deleted instances.

### D — UX Feedback

- **Subcategory settings layout:** Current inline treatment is unintuitive. Preferred: subcategories appear as separate list items, slightly indented under their parent category, each with the same two toggles.
- **Visibility toggles:** Separate Show/Hide and Remote trails toggles are too confusing. Preferred: collapse into a single **Show/Hide** toggle (hides all trails with this category, no matter their origin).

---

## E. Profile — Favourite Sport Migration

| # | Test | Status | Notes |
|---|------|--------|-------|
| E1 | Open the profile page — "favourite sport" section now shows the highest-priority category read-only | ✅ | |
| E2 | The "favourite sport" link leads to Settings → Categories to manage priority | ✅ | |
| E3 | If the account had a favourite sport set before the migration — it now appears as priority 1 in Settings → Categories | ✅ | |

### E — UX Feedback

- **Default top category after migration:** Users who had no favourite sport set end up with Canoeing as priority 1 (first category alphabetically/by seed order). Should default to a more common activity like Hiking instead.

---

## F. Search Visibility & Preferences

| # | Test | Status | Notes |
|---|------|--------|-------|
| F1 | Logged-in user with a hidden category — trails in that category are absent from search results | ✅ | |
| F2 | Make an anonymous (unauthenticated) request to the same search — hidden category filter is NOT applied; trails are visible | ✅ | |
| F3 | Logged-in user with "Remote trails" off for a category — federated trails in that category are absent, local trails remain | ✅ | |

---

## G. Integrations (Strava / Komoot)

| # | Test | Status | Notes |
|---|------|--------|-------|
| G1 | Import a running activity from Strava — trail is assigned the "Running" category | ✅ | |
| G2 | Import a cycling activity from Strava — trail is assigned the "Biking" (or equivalent cycling) category | ✅ | |
| G3 | Import a trail from Komoot with a known activity type — maps to the expected category | ✅ | |

---

## H. Federation & Remote Sync

| # | Test | Status | Notes |
|---|------|--------|-------|
| H1 | Federate/sync a trail that has a category + subcategory — both are matched by normalized name and assigned correctly | ✅ | |
| H2 | Federate a trail with a subcategory that doesn't exist locally — trail is assigned the parent category only; `remote_category`/`remote_subcategory` raw values are persisted | ⚠️ | Works for newly received trails. If the trail already exists, the Update activity category payload is discarded — `TrailFromActivity` (`db/util/activitypub.go:174–184`) early-returns after setting `needs_full_sync=true` for existing records, skipping the category/subcategory parsing block entirely. Category only updates on the next view-triggered full sync. |
| H3 | Federate a trail with a completely unknown category — trail arrives uncategorized but is still visible | ✅ | |
| H4 | Admin adds the missing subcategory after the fact — backfill runs and previously stored trails are assigned the correct subcategory | ✅ | |

### H — Bugs Fixed

- **Nil photo pointer panic when a federated photo download fails:** `TrailFromActivity` (`db/util/activitypub.go:298–308`) originally pre-allocated `photos := make([]*filesystem.File, len(photoURLs))` and assigned each result by index (`photos[i] = photo`). When a download failed and the loop used `continue`, `photos[i]` was left as `nil`. PocketBase panicked on the first nil entry (`i == 0`) when `record.Set("photos", photos)` iterated the slice. Fix: switch to `photos := []*filesystem.File{}` with `photos = append(photos, photo)` so only successfully downloaded files are included and nil entries are never produced.

- **Inbox proxy body re-serialization caused unexpected EOF and broken HTTP signatures:** The SvelteKit inbox handler ([web/src/routes/api/v1/activitypub/user/[handle]/inbox/+server.ts:39](web/src/routes/api/v1/activitypub/user/[handle]/inbox/+server.ts#L39)) read the body with `event.request.json()` then re-serialized it with `JSON.stringify(activity)` before forwarding to Go. Re-serialization produces different bytes than the original (different whitespace/key order), but the original `Content-Length` and `Digest` headers were still forwarded unchanged. The Go handler (`db/routes/activitypub.go:72`) called `io.ReadAll(e.Request.Body)` which hit EOF because the body length didn't match `Content-Length`; had that succeeded, the `Digest` mismatch would have failed HTTP signature verification at line 106. Fix: read the raw body with `event.request.text()`, parse it separately for the actor guard, then forward `bodyText` unchanged so `Content-Length` and `Digest` remain consistent.

---

## I. Deploy / Fresh Instance

| # | Test | Status | Notes |
|---|------|--------|-------|
| I1 | Fresh instance: verify default categories and subcategories (including "Running") are seeded automatically | ✅ | |
| I2 | Existing instance after upgrade: "Running" category and its default subcategories are present | ✅ | |
| I3 | Meilisearch reindex after deploy: `category_id`, `subcategory_id`, `is_federated` fields are present in trail index documents | ✅ | |

slothful-vassal:
Many thanks for the thorough review. I have addressed all findings, but left the “uninstall plugin” behavior (auth fields) unchanged for now. I also updated the category and subcategory settings as suggested.

image

Flomp (approved):

PR #1060, opened 2026-06-15
waypoint editing while drawing a route

slothful-vassal:
- Existing waypoints now remain visible while drawing or editing a route. There is a toggle in the route editing toolbar to show or hide them.
- Right-clicking on the map creates a waypoint at that spot without placing a route anchor, but only when the waypoint markers are visible.
- The trail documentation has been cleaned up and now includes instructions for editing an existing trail.
- The "Stop Drawing" button is available again to protect the trail against accidental changes.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -48,7 +48,7 @@ export default defineConfig({
           label: 'Authentication',
           link: '/use/authentication/'
         }, {
-          label: 'Create a trail',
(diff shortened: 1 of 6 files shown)


Flomp:
Added a small fix:
Clicking on a waypoint while editing a route no longer creates a route anchor and just opens the waypoint popup.

Flomp (approved):

PR #1079, opened 2026-06-28
feat: instance-level ActivityPub federation

Flomp:
## Summary

- Adds an **Application-type instance actor** so each Wanderer instance has its own ActivityPub identity, independent of any user account
- Implements a **mutual Follow lifecycle** between two instance actors: discover → follow → accept/reject → disconnect, with full HTTP-signature delivery and inbox handling
- **Fans out all public content** (trails, comments, summit logs, lists) to connected peer instances via the instance actor's follower/following sets, without touching the existing user-level federation path
- Adds a **federation admin dashboard** at `localhost:8090/federation/` — a standalone SPA injected into the PocketBase admin panel via the UI extension API, matching the PocketBase visual language (IBM Plex Sans, Remix Icons, accent colour theming)
- Fixes **NodeInfo** to count only local public trails (not federated mirrors), and injects the correct version from `web/package.json` via the SvelteKit proxy

## What's new

| Area | Detail |
|---|---|
| Instance actor | `Application`-type AP actor created at startup; RSA keypair generated once and stored encrypted |
| Follow lifecycle | Outbound Follow → remote Accept/Reject → local status update; inbound Follow → admin Approve/Reject → Accept/Reject delivery |
| Content fanout | `instanceFollowerInboxes()` appended to recipient list for all Create/Update/Delete activities on public content |
| Instance inbox | `POST /api/v1/activitypub/instance/inbox` — HTTP-sig verified, dispatches Follow/Accept/Reject/Undo/Create/Update/Delete |
| Admin API | `POST /federation/discover`, `/follow`, `/approve/:id`, `/reject/:id`, `/disconnect/:id`, `GET /federation/peers` |
| Admin UI | Alpine.js SPA at `GET /federation/` — discover, connect, manage peers, confirmation modals, status badges |
| NodeInfo fix | Trail count filtered to `is_local = 1`; version injected from `package.json` in SvelteKit proxy |
| Re-follow after rejection | Rejected follow record is cleared so admins can retry without a 409 |

### Instance actor

- [ ] Fresh database: instance actor record is created automatically at startup with `actor_type = instance`, `is_local = true`, and a non-empty RSA public key
- [ ] Restart the server: the same actor record is reused (keypair is NOT regenerated)
- [ ] `GET /api/v1/activitypub/instance` returns a valid ActivityPub `Application` object with `inbox`, `outbox`, and `publicKey`

### Discover

- [ ] `/federation/` → paste a valid remote Wanderer URL → "Discover" shows the remote instance domain, version, user count, and trail count
- [ ] Paste the local instance URL → discover returns an error (self-follow guard)
- [ ] Paste a non-Wanderer URL → discover returns an appropriate error
- [ ] Discover button is disabled when the URL field is empty

### Follow lifecycle

- [ ] After discover, click "Connect" → a `pending` follow record appears in the peers table with "Awaiting approval…" and a "Cancel request" button
- [ ] Clicking "Cancel request" opens the confirmation modal ("Cancel follow request?") and on confirm withdraws the request; the row disappears
- [ ] On the remote instance, the inbound follow request appears as `pending` with Approve / Reject buttons
- [ ] Approve → both sides show `accepted`; content starts flowing
- [ ] Reject → requesting instance sees `rejected`; re-attempting the follow succeeds (rejected record is cleared first)
- [ ] Disconnect (accepted peer) → opens "Disconnect instance?" modal → on confirm the row is removed from both sides

### Content sync

- [ ] Create a public trail on instance A → it appears on instance B within seconds
- [ ] Update the trail on A → the change propagates to B
- [ ] Delete the trail on A → it is removed from B
- [ ] A private trail (`is_public = false`) is never delivered to B
- [ ] Comments, summit logs, and lists follow the same Create/Update/Delete flow

### Admin UI

- [ ] The PocketBase admin header shows a "Federation" link that opens `/federation/` in a new tab
- [ ] Closing the tab via the "Close" button (top-right) works; it does not navigate away
- [ ] Visiting `/federation/` without a superuser session shows the login prompt card, not the peer table
- [ ] Dark mode and non-default accent colours are picked up correctly on first load (no flash)
- [ ] Empty state ("No peer connections yet") is centred in the table

### NodeInfo

- [ ] `GET /.well-known/nodeinfo/2.1` → `software.version` matches the value in `web/package.json`
- [ ] `localPosts` count equals the number of public trails authored on this instance only (federated trails excluded)


slothful-vassal (changes requested):
Thanks for putting this together. Having a dedicated instance actor looks like a useful foundation for connecting Wanderer instances. After looking through the implementation, there are a few things I think we should address before broader testing:

1. **Internal proxy protection — P1**

   The new instance inbox trusts `X-Forwarded-Path` without checking the internal proxy secret or restricting the recipient to the instance actor. If the backend is directly reachable, a validly signed request can therefore target a user inbox through this endpoint, bypassing the existing internal proxy protection.

2. **Create deduplication can drop personal feed entries — P1**

   In `db/federation/create.go`, the new deduplication returns immediately when a trail or list already exists. However, inserting the recipient’s feed entry happens after that check.

   If an activity reaches the instance inbox first and a user inbox afterwards, the user delivery is discarded and their personal feed entry is never created. The same problem affects multiple local users following the same remote author.

   Could we separate object deduplication from recipient-specific processing? Repeated delivery should avoid duplicate content while still ensuring that each intended recipient gets their feed entry.

3. **An old outbound relationship can hide a new incoming request — P2**

   `buildPeerEntries` in `db/routes/federation_admin.go` prefers an existing outbound entry even when it is rejected and a new inbound request is waiting for approval.

   For example:

   - A requests a connection to B.
   - B rejects it.
   - B later requests a connection to A.

   A now has an old `outbound/rejected` record and a new `inbound/pending` record, but the dashboard only shows the rejected one. The incoming request cannot be approved through the UI.

   The same issue can occur when both instances initiate a connection at roughly the same time. Pending inbound requests need to remain visible and actionable independently of the outbound state.

4. **Disconnect does not fully disconnect a mutual peer — P2**

   When both directions have accepted Follow records, the peer list combines them into one “mutual” entry using the outbound Follow ID.

   `FederationDisconnect` then deletes only that outbound record. The accepted inbound record remains, and `instanceFollowerInboxes` continues to use it for delivery. Synchronization can therefore continue even though the confirmation dialog says it will stop.

   Disconnecting the displayed peer should terminate both active relationships and send the corresponding activities.

5. **Retrying after rejection depends on delivery order — P2**

   When retrying a rejected connection, `createOutboundFollow` deletes the old record and creates a new one. The hooks send the old `Undo` and the new `Follow` asynchronously.

   If the new Follow arrives first, `ProcessFollowActivity` treats it as a duplicate because the rejected relationship still exists. The delayed Undo then removes that relationship. The requesting instance remains `pending`, while the receiving instance has no request left to approve.

   It would help to associate lifecycle processing with the specific Follow activity, so a delayed message from an earlier attempt cannot discard a new connection attempt.

6. **The instance actor is included in profile search during startup — P2**

   The actor hooks skip instance actors during incremental indexing, but `initMeilisearchDocuments` in `db/main.go` still loads and indexes every actor.

   This includes the instance actor in the normal actor search index whenever the index is rebuilt. The exclusion needs to apply to the full indexing path as well.

There is also one behavior I would like us to clarify: should approving or disconnecting a peer control the acceptance of incoming instance content, or only outgoing delivery? Currently, the instance inbox accepts validly signed content without an accepted peer relationship. If approval is intended to control incoming synchronization too, that check is missing. It would need to account for content being signed by individual user actors.

The merge conflicts also deserve some care, particularly in `db/util/safe_fetch.go`: the shared client factory needs to retain the existing SSRF protections, including the `::/96` block. In `db/main.go`, both instance initialization and subcategory initialization need to be preserved.

PR #1081, opened 2026-06-29
change trail edit in draw mode

slothful-vassal:
## Summary
- The trail edit page now auto-starts paint/drawing mode only for: existing trails that are not completed or duplicated trails
- Completed trails now open the edit page without entering drawing mode automatically.
- The initial map zoom now accounts for the visible elevation profile at the bottom, so the route is not hidden behind the chart.
- Fixed elevation profile toggling so manually hiding it while drawing stays respected across subsequent route/anchor updates.

## Tests

- [x] Edit a completed trail: Drawing mode should not be active automatically.
- [x] Edit a not completed trail: Drawing mode should be active automatically.
- [x] Duplicate a trail: Drawing mode should be active automatically.
- [x] The route should be visible and not hidden behind the elevation chart.
- [x] Manually clicking “Edit route” should still activate drawing mode.
- [x] Create a new trail without duplication: Drawing mode should not auto-start.
- [x] While drawing/editing a route, manually hide the elevation profile, then add another anchor point; the elevation profile should remain hidden until manually shown again.


Flomp:
> Edit a completed trail: Drawing mode should not be active automatically.

Pass

> Duplicate a trail: Drawing mode should be active automatically.

Pass

> The route should be visible and not hidden behind the elevation chart.

Pass

> Manually clicking “Edit route” should still activate drawing mode.

Pass

> Create a new trail without duplication: Drawing mode should not auto-start.

Pass

> While drawing/editing a route, manually hide the elevation profile, then add another anchor point; the elevation profile should remain hidden until manually shown again.

Pass

Flomp (approved):

PR #1094, opened 2026-07-03
Add Intervals.icu plugin integration

palhaland:
As Intervals.icu is another great aggregator for Garmin data (in addition to Strava and many other sources). Which can give better analysis of training data, and you can ask coaches for help.


slothful-vassal:
Hello,

Thank you very much for your contribution. We currently have no plans to integrate any more Trail plugins into our repository. However, we would be grateful if you could make your plugin available to other users via your own repository. Once the plugin system has been rolled out, we will publish a list of external plugins to which you can add yours. We ask for your patience for a little while longer.

Best regards

PR #1103, opened 2026-07-06
fix: fall back to cached trail/list when remote actor fetch fails

Flomp:
## Summary
- `findLocalTrailByRemoteInfo` (`db/routes/remote_trail.go`) and `findLocalListByRemoteInfo` (`db/routes/remote_list.go`) discarded a previously-cached actor record whenever the live actor lookup returned an error (e.g. the remote instance is unreachable), turning what should be a graceful fallback to cached content into a 500.
- Both functions now use a non-nil (possibly stale) cached actor, or fall back to a plain local ID lookup, before giving up — mirroring the existing pattern in `ActivitypubActorGet`.
- `RemoteListGet` was missing the `ErrProfilePrivate` 404 handling that `RemoteTrailGet` already had; added it, and fixed a copy-pasted "Failed to resolve trail" error message.

## Test plan
- [ ] Access a trail by an actor of an instance that is no longer available. The trail loads without showing an error.
- [ ] Access a list by an actor of an instance that is no longer available. The list loads without showing an error.


slothful-vassal (commented):
I can still reproduce a 500 when a trail already exists locally, is marked as `needs_full_sync`, and the remote instance is unavailable. Could we fall back to the locally cached trail if it was successfully synced before?

Flomp:
Good catch, fixed in the latest push.

The `needs_full_sync` path forced a blocking sync and turned any sync error into a 500. It now falls back to the local copy when the origin never actually got to answer:

- Falls back: transport failure (instance down/DNS/timeout), remote 5xx, or our own outbound rate limiter refusing before the request leaves the instance.
- Does not fall back: a remote 4xx. That's the origin answering. The trail was deleted or is no longer public. So serving stale cache there would re-publish withdrawn content.

slothful-vassal (changes requested):
### Repro: never-synced list placeholder is served as "0 trails" with HTTP 200

1. alice@A follows bob@B.
2. bob@B creates a new public list with two public trails (after the follow, so A receives the Create and stores a placeholder: `needs_full_sync = true`, no trails).
3. Do not open the list on A. Take B offline (stopped process or 502 from the tunnel, both count as "remote unavailable").
4. alice@A opens the list from her feed.

Actual: `GET /api/v1/list/<id>?handle=@bob@<B-host>&expand=…` → 200, `trails: []`, `needs_full_sync: true`; the UI shows the list with 0 trails as if that were its content; log says `serving cached list after failed remote sync`.

Expected: the fallback should only serve copies that have been fully synced at least once. A placeholder has no cached content, so this case should behave as before the PR (error) or the response must be marked as incomplete so the UI can say so. Serving a previously synced list from cache while B is offline is fine and is what the PR is for.

Counter-check: bring B back, open again → full sync runs, list complete. Take B offline once more and open the list a third time → now the cached copy is legitimately served with both trails.

Cause: `cachedRecordFallback` only requires a stored id; `needs_full_sync` already tells a placeholder from a synced copy but is not consulted. Same path exists for trail placeholders (Create/Announce).

Flomp:
Ah, sneaky.

`needs_full_sync` alone can't fix it: the fallback only runs inside the `needs_full_sync == true` branch, and a previously synced copy re-flagged by an Update carries the same value as a never-synced placeholder. Checking it would disable the fallback entirely, including the case from your first comment.

So there's a new bool `full_sync_completed` on trails and lists, set only when a full sync finishes. The fallback requires it. A placeholder with the origin down now errors as before the PR; a previously synced copy is served from cache. The migration backfills existing remote records that aren't flagged stale, so installs upgrading don't lose the fallback until their next successful sync.

slothful-vassal (approved):

PR #948, opened 2026-04-25
Add asset plugin type and first-party Immich plugin

slothful-vassal:
## Summary

This PR introduces **asset plugins**, a new plugin category that connects wanderer to external media libraries, and ships **Immich** as the first integration: Immich suggests geotagged photos that match a trail or waypoint, and wanderer imports or links them as photo assets. This implements the feature requested in several issues (#160, #395). The plugin type is generic, so other libraries (e.g. PhotoPrism, #530) can follow without further core changes.

### Data model & migration

- New `assets` collection plus link collections for trails, waypoints, and summit logs, with access rules, APIs, and hooks.
- Existing photos are migrated into asset records, and the legacy `photos` fields are removed from the trails, waypoints, and summit_logs collections — assets are now the single source of truth.
- Backward compatibility: API responses still expose a `photos` array, now computed from asset links, and federation/full-sync converts legacy `photos` file URLs from older instances into assets.

### Plugin system

- New plugin `type: "assets"` with an `asset_library.v1` capability; installed plugins are now categorized as `trails` or `assets`.
- Host-owned configured connectors, asset import limits, and thumbnail/import/materialize/repair/delete flows. Plugin schema, SDK, and docs updated accordingly.
- Two storage modes per asset plugin (`photoMode`): photos are either **copied** into wanderer (`copy`), or kept in the source library and referenced (`link_private`). Linked photos are materialized (copied in) before their target becomes public, and a guard prevents disabling a plugin while linked photos still depend on it.
- Because any public or federated trail always materializes its photos, `link_private` only saves storage for trails that stay private. The settings UI states this so the tradeoff isn't surprising.

### Immich plugin

- Matches photos against the trail by capture time and GPS location; supports preview or original quality imports and an "own photos only" mode.
- Uses the generic storage modes above; defaults to `copy`.
- Least-privilege API key model — see upgrade notes below.

### UI

- Photo search and selection from the media library in the trail editor, waypoint attachment, and waypoint creation directly from photos.
- Asset plugin settings, including admin actions to materialize, repair, or delete remote-linked photos in bulk (also triggered when disabling a plugin). These run asynchronously on the server; the settings page polls the job and shows a progress bar.
- New data maintenance pages: duplicates, photo metadata backfill, trails without photos, and orphaned photos.

### Existing flows

[...]

PR #1134, opened 2026-07-29
Fix duplicated shared-list rendering

### LINKED ISSUE #1128: Shared list page has double sidebar
### What happened?

If I click on a shared list, e.g. https://wanderer.matteofortini.it/lists/@mfortini@localhost/cjtjvp7bler15bw I get a double sidebar instead of the right page (see screenshot)

Image

### Version

v0.20.0

snowyukitty:
Fixes #1128.

## What changed

- Build long collapsed descriptions as escaped plain text instead of injecting
  a raw HTML string cut at character 100.
- Count the first 100 Unicode code points of the formatted plain-text preview
  and decode supported HTML entities exactly once.
- Render the complete existing rich text only when the description does not
  need a preview or after the user selects **Read more**.
- Add an end-to-end regression that creates bold and italic content whose raw
  HTML is cut inside both inline elements, opens the shared-list route, and
  checks the resulting page structure and preview.

## Why

The previous collapsed state used `description.substring(0, 100)` inside
`{@html}`. When that cutoff lands inside an inline element, the browser repairs
the incomplete SSR markup differently from the client render.

On the exact pre-change `main` revision, a description cut inside nested
`<strong>` and `<em>` elements reproduced the reported failure in three out of
three runs:

- `main`: 2
- `.list-list`: 2
- `#trail-map`: 2

That reproduction does not emit `hydration_mismatch`, so the regression asserts
the page-section counts directly rather than relying only on console output.
The candidate produced one of each section in every focused run.

## How it was tested

- Exact-base regression: reproduced the duplicate layout in 3/3 runs.
- Candidate focused regression: passed 3/3 before mutation testing.
- Final focused `--repeat-each=10 --retries=0`: 10/10 passed.
- Mutation checks detected:
  - restoration of raw truncated HTML;
  - a **Read more** control moved into an overflow-hidden container;
  - the existing 20 px first-paragraph spacing regression;
  - removal of preview tag stripping;
  - an inline-markup fallback to raw truncated HTML;
  - chained entity replacements that decode text twice; and
  - UTF-16 slicing that splits an astral character at the cutoff.
- `npm run check`: 0 errors, 0 warnings.
- Vitest: 3/3 passed.
- `npm run build`: passed.
- `git diff --check`: passed.
- No-cache web and database image builds: passed, with running container image
  IDs verified.

The repository's existing `shows a list card` Playwright case timed out at
`lists_page.ts:65` while waiting 10 seconds for `.list-list-item`. The same
timeout reproduced on the untouched base and passed on retry. The full
Playwright project completed with 8 passed and 1 flaky. The new regression
passed 10/10 with retries disabled and deletes its exact record by ID.

## Known limitations

- A collapsed long description is intentionally plain text. Links and inline
  formatting become active only after **Read more** is selected.
- The former 96 px `max-h-24 overflow-hidden` cap is removed. Long previews are
  bounded to 100 Unicode code points; descriptions at or below that length
  render their complete rich text and can therefore be taller than before
  instead of being silently clipped.
- This patch is intentionally atomic to
  `components/list/list_panel.svelte`. Similar raw-rich-text truncation remains
  in `routes/profile/[handle]/+page.svelte`,
  `waypoint/waypoint_card.svelte`, `trail/trail_timeline.svelte`, and
  `trail/trail_info_panel.svelte`. Those sites need separate reproduction and
  collision review; `trail_info_panel.svelte` is currently also modified by
  open PR #948.


Flomp:
Thanks a lot for this.
I've built on top of your commit rather than merging it as-is, for two reasons:

**1. We already had a helper for this.** `formatHTMLAsText` in `format_util.ts` does the HTML-to-text conversion, and a couple of components already use it for previews. I've made that helper work on both the server and the client instead. 

**2. The same bug exists in three other places.** Trail descriptions, activity cards and feed cards all truncate rich text the same way. All four now go through one shared helper, so this can't drift apart again.

Flomp (dismissed):

Flomp (approved):

PR #1159, opened 2026-08-10
fix(db): keep the OAuth2 provider username on signup

mfuhrmann:
Fixes root cause 2 of #1026.

PocketBase only maps the provider username when the raw value already satisfies `users.username` (`^[\w][\w.\-]*$`, min 3). OSM display names contain spaces and umlauts, so accounts end up with a generated `usersNNNNNN` name.

This transliterates accented characters (`Karl Dörfinger` → `Karl_Doerfinger`), replaces anything left over with underscores, and resolves collisions with a numeric suffix. When nothing usable remains the username is left unset and PocketBase generates one as before.

Assigning the raw value instead is not an option: record validation rejects it, so signup fails outright.

Tested with OSM logins on a self-hosted instance; unit tests in `db/util/username_test.go`.

### DIFF
--- a/db/hooks/users.go
+++ b/db/hooks/users.go
@@ -71,6 +71,45 @@ func UpdateUserHandler(client meilisearch.ServiceManager) func(e *core.RecordEve
 	}
 }
 
+// OAuth2UsernameHandler assigns the username reported by the OAuth2 provider
+// to accounts created through OAuth2.
+//
+// PocketBase maps the provider username itself, but only when the raw value
+// already satisfies the users.username field. Providers whose usernames are
+// display names - OpenStreetMap, for instance - therefore end up with a
+// generated "usersNNNNNN" name. Sanitising the value first keeps the name the
+// user signed up with recognisable.
+func OAuth2UsernameHandler() func(e *core.RecordAuthWithOAuth2RequestEvent) error {
+	return func(e *core.RecordAuthWithOAuth2RequestEvent) error {
+		if !e.IsNewRecord || e.OAuth2User == nil {
+			return e.Next()
+		}
+
+		// a username submitted by the client takes precedence
+		if submitted, _ := e.CreateData["username"].(string); submitted != "" {
+			return e.Next()
+		}
+
+		username := util.SanitizeUsername(e.OAuth2User.Username)
+		if username == "" {
+			return e.Next()
+		}
+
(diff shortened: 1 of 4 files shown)


slothful-vassal (changes requested):
Thanks for following up with this PR! Sanitizing provider names before assigning them makes sense, and keeping existing accounts unchanged is a good choice.

I checked the signup flow a bit more closely and found a few edge cases worth addressing before merging:

1. **Case-insensitive username collisions**  
   `UniqueUsername()` uses a case-sensitive lookup, while the username’s unique index uses `COLLATE NOCASE`. If `Alice` already exists, the helper considers `alice` available, but saving the account fails. Without the hook, PocketBase falls back to a generated username and signup succeeds. Could the lookup follow the same comparison rules as the index?

2. **Simultaneous signups choosing the same name**  
   The lookup happens before PocketBase’s account-creation transaction. Two requests can therefore select the same available name, causing one signup to fail. Setting an explicit `CreateData["username"]` also bypasses PocketBase’s usual fallback. Could we move the allocation into the creation transaction, or handle collisions with a retry/fallback?

3. **Respecting the configured username mapping**  
   The hook always writes `CreateData["username"]`, even when an admin has disabled the OAuth username mapping or mapped it elsewhere. For example, disabling the mapping currently produces a generated name, but with this hook the provider name is still assigned. It would be good to preserve that configuration choice.

4. **Respecting the actual username field rules**  
   The sanitizer assumes the default pattern and 3–150 character limits. If an installation has a shorter maximum, it can return a name that fails validation. PocketBase previously handled this by generating a name instead. Could we validate against the actual collection field before assigning the result?

mfuhrmann:
@slothful-vassal Thanks also here for the review :hugs:  You were right on all four points. I changed the approach:

- No more `CreateData["username"]`. The hook now only cleans up the provider's username. PocketBase then checks it and saves it in the same step. If the name is taken by then, it falls back to a generated name, so signup no longer fails. 
- The free-name check now compares like the unique index does, so `alice` counts as taken when `Alice` exists.
- The hook does nothing when the mapping is off or points to another field, or when the client sent a username.
- min, max and pattern come from the real `username` field, not hard-coded values.

I added tests for each case.

slothful-vassal (approved):
Thanks for both PRs and for addressing the review feedback! I tested signup with Authentik. Looks good to me!

PR #1166, opened 2026-08-18
plugin info modal added

slothful-vassal:
Hi,

To give users clearer information about a plugin’s purpose, availability, and installed version, I’ve added an info modal to the plugin settings.

Plugins can provide localized information through their manifest metadata. Plugins based on open-source projects can also include an optional donation link. In addition, a link to the provider's homepage can be provided.

image

image

image


Flomp:
Thanks! Looks good. I made some small changes:

**1. Guard non-string localized values.** `localizedTextForCandidates` calls `text.trim()` on values only *typed* as `string`. Every map this PR touches is now filtered by `localizedMetadata`, but `providerCategoryLabel` still casts raw manifest data (`localizedText(labels as LocalizedTextMap, …)`). A manifest with `providerCategories.hiking.labels = {"en": 42}` throws a TypeError when the instance settings modal renders its category dropdown. This predates the PR, but the locale-key normalization slightly widens it (`{"EN": 42}` now throws too), and it contradicts the README's new promise that "invalid localized values are ignored". Added a `typeof` guard.

**2. Keep the info modal to setup errors.** `error={pluginCardError(infoPlugin, instanceForPlugin(infoPlugin))}` falls through to `instanceError(instance)` when `status === "available"`, so a correctly installed plugin that merely hit a sync failure renders "Availability: **Available**" directly above a red "Provider rate limit reached. Try again later." I extracted a `pluginSetupError()` helper that returns `""` unless the status is actually `error`. 

**3. Restored the Norwegian `startDate` description.** The rewrite of the Hammerhead config field dropped the existing `"no"` entry. Was that intentional? 

**4. Documented the redaction in the OpenAPI block.** Added a `description` on the field saying so.

Flomp (approved):

slothful-vassal (commented):
Many thanks for the fixes.

Yes, I removed the Norwegian description intentionally because the German and English text had changed, and I didn’t want to keep an outdated translation. But you’re right—I could have used DeepL to provide an updated Norwegian version. Thanks for doing that.

PR #1167, opened 2026-08-19
Show waypoint photo thumbnails on the map and elevation profile.

### LINKED ISSUE #454: Photos Not shown at waypoints in the map
image

mfortini:
## Summary

Waypoint markers on the map and elevation profile now show photo/video previews, and focus is synced between map, profile, and the waypoint section on the page.

Previously map waypoints were icon-only placeholders; elevation markers had no media preview. This makes it easier to identify photo waypoints while browsing a trail (related to #454).

### Intended interactions

- Click a map waypoint: exclusive popup with up to 3 media thumbnails, name, and distance; elevation marker gets a highlight ring; page scrolls to that waypoint (gallery does not open).
- Click elevation marker: focuses the matching map popup / page anchor; hover shows a thumbnail above the icon.
- Media for saved photos uses PocketBase thumbs; local unsaved `_photos` use object URLs that are revoked on cleanup.

### How tested

- Unit: `waypoint_map_util` (anchor ids, media URLs/thumbs, revoke, highlight).
- Manual: trail map + list map; open/close several popups and confirm network only loads on open; imperial unit distance; replace waypoint photo in edit and confirm profile marker updates; confirm map open does not force elevation thumbnail visible.

### Known limitations / follow-ups

- Keyboard/touch affordances for profile markers not expanded yet.
- Popups with many photos are capped (first 3); no in-popup scroll.
- Elevation hover thumbnail positioning on very small screens may need polish.

May fix #454

https://github.com/user-attachments/assets/0c765290-d72b-47b7-94fb-4f37ab78e823


web/src/lib/util/maplibre_util.ts
```diff
@@ -40,32 +45,68 @@ export function createMarkerFromWaypoint(waypoint: Waypoint, onDragEnd?: (marker
     })
 
     const content = document.createElement("div");
-    content.className = "p-2"
+    content.className = "p-2 waypoint-popup"
+
+    const media = getWaypointPopupMedia(waypoint);
```
slothful-vassal: **Blocking**: Assigning src while constructing every marker starts requests for all original waypoint media, even while the popup is still closed. This is reproducible easily. Could we create and load the preview only when the popup is opened, and limit it to a small number of media items?

web/src/lib/vendor/maplibre-elevation-profile/elevationprofile.ts
```diff
@@ -906,6 +884,90 @@ export class ElevationProfile {
...
+    private syncWaypointMarkers(chart: {
+        scales: Record<string, { getPixelForValue: (value: number) => number }>;
+    }) {
+        const waypointContainer = document.getElementById("waypoint-container");
+        if (!waypointContainer) {
+            return;
+        }
+
+        const xScale = chart.scales.x;
+        const signature = this.waypoints
```
slothful-vassal, line 899: **Blocking**: The cache signature only contains the waypoint ID and saved-photo count. Replacing a waypoint under the same ID retains the old DOM and click closure, so coordinates, icon, name, thumbnail, or local _photos can remain stale. Please invalidate or update the marker using the current waypoint data.

web/src/lib/util/waypoint_map_util.ts
```diff
@@ -0,0 +1,108 @@
...
+    const urls: string[] = [];
+
+    for (const photo of waypoint.photos ?? []) {
+        const url = getFileURL(waypoint, photo);
+        if (url) {
+            urls.push(url);
+        }
+    }
+
+    for (const file of waypoint._photos ?? []) {
```
slothful-vassal, line 55: **Blocking**: These Object URLs are never revoked when map markers or profile markers are rebuilt or removed. Since the helper only returns the URL string, callers also have no clear cleanup mechanism. Could we define ownership and revoke the URLs during marker/profile cleanup?

web/src/lib/util/maplibre_util.ts
```diff
@@ -82,6 +123,24 @@ export function createMarkerFromWaypoint(waypoint: Waypoint, onDragEnd?: (marker
...
 
+let exclusiveWaypointPopup: M.Popup | undefined;
+
+function bindExclusivePopup(popup: M.Popup, waypoint: Waypoint) {
+    popup.on("open", () => {
+        if (exclusiveWaypointPopup && exclusiveWaypointPopup !== popup) {
+            exclusiveWaypointPopup.remove();
+        }
+        exclusiveWaypointPopup = popup;
+        highlightElevationWaypoint(waypoint.id);
```
slothful-vassal, line 179: **Blocking**: Opening the map popup immediately activates the elevation marker, and the active CSS class also displays its thumbnail. As a result, clicking one map marker opens both the map popup and the elevation preview. If only highlighting is intended, the highlight and preview visibility should use separate states; alternatively, the elevation preview should only be activated by the explicit popup-content interaction.

web/src/lib/util/maplibre_util.ts
```diff
@@ -40,32 +45,68 @@ export function createMarkerFromWaypoint(waypoint: Waypoint, onDragEnd?: (marker
...
-    nameElement.textContent = waypoint.name ?? "-";
     if (waypoint.name?.length) {
-        nameElement.classList.add("ml-2")
+        const nameElement = document.createElement("b");
+        nameElement.className = "block";
+        nameElement.textContent = waypoint.name;
+        content.appendChild(nameElement);
+    }
+
+    if (typeof waypoint.distance_from_start === "number") {
```
slothful-vassal: **Blocking** correctness issue: The popup snapshots distance_from_start before the elevation profile recalculates it, so new or moved waypoints can show a missing or stale value. The value is also incorrect in imperial mode because an already mile-adjusted distance is later treated as meters.

slothful-vassal (changes requested):
Thanks for working on this — showing waypoint photos on the map is a useful improvement.

Could you please expand the PR description in line with our [contribution guidelines](https://github.com/open-wanderer/wanderer/blob/dev/CONTRIBUTING.md#describe-the-change-clearly)? In particular, it would help to describe the intended interactions, how they were tested, and any known limitations. A screenshot or short recording would also be useful for this UI change.

I left blocking inline comments concerning:
- eager loading of original media
- stale elevation-marker state after waypoint edits
- Object URL cleanup
- both map and elevation overlays opening together
- distance correctness

The following would be welcome as follow-ups and do not necessarily need to block this PR:
- keyboard and touch interaction
- limiting or scrolling popups containing many photos
- responsive positioning of elevation thumbnails

Thanks again for the contribution!

slothful-vassal (changes requested):
Thanks for the update! I found a few remaining issues:

- **Profile marker alignment:** With at least two waypoints without photos, the markers become vertically staggered.
- **Thumbnail configuration:** `600x0` and `100x0` aren’t configured for `waypoints.photos`. PocketBase therefore serves the original images; a migration is needed.
- **Replacing local photos:** Replacing a photo while keeping the same `_photos` count leaves the old profile preview cached.
- **Map marker navigation:** Scrolling to the waypoint currently requires a second click inside the popup. The PR description specifies clicking the marker.

PR #1168, opened 2026-08-19
Draw list routes from indexed polylines instead of downloading GPX files

mfortini:
The lists overview map is empty or very slow because it required downloading every trail GPX sequentially before rendering.

- Index each list's bounding box and trail polylines in Meilisearch
- Render simplified preview routes on the overview map from polylines
- Fit the viewport to the combined trail bounds with padding
- Load full GPX only when a single trail is opened inside a list


db/util/meilisearch.go
```diff
@@ -224,6 +301,56 @@ func documentFromListRecord(r *core.Record, author *core.Record, includeShares b
 		"iri":            r.GetString("iri"),
 	}
 
+	expandedTrails := r.ExpandedAll("trails")
```
slothful-vassal: **Blocking**: this can expose private trail geometry through a public list.
expandedTrails is populated server-side and is not scoped to the current viewer. The code below derives the list center, bounds, and polylines from every expanded trail without checking the trail’s visibility.
Anonymous Meilisearch tokens only restrict the lists index to public = true (db/routes/search_token.go:13-16); they do not reapply the referenced trails’ permissions. At the same time, the list editor explicitly allows making a list public while keeping its trails private (web/src/routes/lists/edit/[id]/+page.svelte, the keep-private path).
Reproduction:
1. Create a private trail and add it to a list.
2. Make the list public and choose to keep the trails private.
3. Log out.
4. Search for the list through POST /api/v1/search/lists.
5. The result still contains trail_polylines, lat/lon, and min_*/max_* derived from the private trail.
Filtering only public trails while indexing would not fully solve this either: the owner, shared users, and anonymous users can have different access to the trails, while the list has only one shared Meilisearch document. The geometry therefore needs to remain behind trail-level permissions.

slothful-vassal (changes requested):
Thanks for working on this — I agree with the goal of making the list view load quickly without pulling every GPX file up front.

Up front: I don't think this indexing approach should be merged as it stands, mainly because of trail permissions. A public list can currently expose the geometry and bounds of private trails it contains. I left a blocking inline comment on the geometry aggregation with the concrete code path and reproduction. And that isn't something tighter sync logic can fix — it follows from list documents owning trail geometry in the first place.

Related issues from the same root:

* Trail updates, deletions, and visibility changes don't reliably propagate to the derived list geometry, and removed geometry can linger in Meilisearch.
* The per-list polyline limit doesn't respect the global map limit, so the total amount of geometry can grow well past it across several lists.

What I'd suggest instead: keep geometry in the trail index and compose list previews server-side — resolve the trails visible in each list, batch-fetch their geometry from the trail index, enforce permissions in one place, deduplicate trails shared across lists, and apply a single global polyline budget. Meilisearch can't join the two indices itself, but a bounded server-side composition avoids both the duplication and N+1 requests, and leaves room for caching later if measurements show it's needed.

For what it's worth, I ran into the same bottleneck a while back in #555 and approached it from the lazy-loading side instead. I never finished it, but it's a reminder that this has been a real problem for a while.

There are also smaller implementation issues — the frontend typecheck is currently failing, plus a few map interaction and state-management regressions. Worth addressing in the rework, but secondary to the ownership and permission questions.

The UI work here is worth keeping either way. Up to you whether you adapt this PR or open a fresh one and carry it over. Happy to sketch out the endpoint shape or look at an early draft if that would help.

slothful-vassal (changes requested):
Thank you for revising this and taking the earlier feedback on board. Keeping the geometry in the trail index and composing the previews under the viewer's token resolves the main concern regarding indexing that I raised in my previous review. This looks like a good approach.

However, I found the following remaining correctness issues that I’d like to resolve before merging:

- **Trails disappear when they exceed the preview limit.** They should fall back to markers, but the endpoint reads `lat`/`lon` while the index stores coordinates in `_geo.lat`/`_geo.lng`. Once their polyline is removed, there is no usable position for a marker.

- **Opening a summit-log entry’s GPX on the map no longer fits the full track.** The shared map component now uses a fixed area around the track’s starting point instead of its available GPX bounds. This is a side effect of the changes to the map component used by the list view.

- **Changing filters can leave the map at the previous location.** For example, switching from one route near Zurich to one near Tokyo loads the Tokyo route but leaves the viewport on Zurich. The fit key is updated before the new geometry arrives, and the unchanged route count prevents another fit.

- **A late fallback response can overwrite the correct viewport.** If the overview preview arrives before the global bounding-box response, the map first fits the list routes correctly and then jumps to the global bounds. The fallback needs to check whether its result is still relevant after the request completes.

- **Opening an empty list shows routes from other lists.** When `expand.trails` is absent, the selected list falls back to `overviewTrails`. In that case, it should use an empty array.

- **Dragging the map can open a list unintentionally.** The new navigation uses `mouseup`, so dragging from a preview line and releasing the mouse can select its list. A click handler or explicit drag check should prevent this.

- **Returning to the overview can accumulate click handlers.** After updating the overview, opening a list, and returning, one click can trigger multiple list requests. An existing handler-reference problem combines with the new cleanup to leave the original handler registered while adding another. The same reproduction produces one callback on `dev` and two on this branch.

- **Some federated lists have no preview despite their trails being available locally.** The new fallback copies remote trail IDs, but imported trails have different local IDs. Those references need to be resolved through their IRIs before querying the local index.

Separately, I’d like to discuss **trails shared by multiple lists**. Suppose the limit is **2**, and two lists both contain **A and B**. Both trails are allowed by the budget, but each is drawn once per list: **A, B, A, B — four line features**.

This preserves the colour and click target of each list. However, identical routes overlap exactly and, since the click handler only uses the first matching feature, users cannot reliably choose between their lists.

One possible approach would be to draw each route once while retaining all its list memberships. Clicking a route belonging to one list could open it directly; clicking a route belonging to several lists could offer a small list chooser. That would also make the preview limit correspond to the number of routes actually drawn.

I’m open to other approaches here – the important things are to preserve list membership, make shared routes selectable and ensure the preview limit has a clear meaning.

PR #1169, opened 2026-08-19
Give lists persistent /lists/{handle}/{id} URLs

mfortini:
Lists now use stable URLs with handle and id so that refreshing, sharing, or navigating back preserves the selected list and trail. The page loader fetches the selected list server-side and the search bar on /map links directly to the canonical list URL.


slothful-vassal (changes requested):
Thanks for this — the persistent URLs are a real improvement, and the '/map' search link is perfect. While testing, I came across three issues that I'd like to see addressed before merging:

1. Infinite scroll drops entries

Repro: On the /lists page, scroll until page 2 loads (60 entries). Open a list from the second page, then use the browser back button to return to /lists and keep scrolling.

The result is that the 30 entries from page 2 disappear and the sidebar jumps from page 1 straight to page 3.

2. The map no longer zooms out when deselecting.

Repro: Open a list — the map fits the list's BBOX correctly. Press Back.

The result is that the map stays on the previous list's box while the sidebar shows the global list index.

3. The in-app Back button and the browser Back button behave differently.

Repro: From /lists, open a list, then press the in-app Back arrow and then the browser Back button.

Result: you will return to the list you just left.

wendelinism:
Does this commit also preserve the view selection i.e. from "Lists" (let's say public lists were de-selected), then switching to "Trails" and then back to "Lists"? 
I was going to add a feature request on making view selections persistent, but possibly this is already covered here?!

PR #1171, opened 2026-08-24
Re-add Further Heights to the server list

mozillamonks:
Re-adding furtherheights.com (originally shared in discussion #361, image already in server/imgs/). The instance is on wanderer 0.20.0.


slothful-vassal (dismissed):

slothful-vassal (approved):

slothful-vassal:
Hello,

We’re sorry that we removed your entry. Was your server temporarily unavailable? We’d be happy to add it back to the list. Thank you very much.

mozillamonks:
We had some DNS issues were www. was serving but the bare domain wasn't; should be fully resolved now and I have monitors set up to keep a closer eye on it.

PR #1172, opened 2026-08-24
fix(web): preserve share token in fullscreen map

### LINKED ISSUE #1136: Big map doesn't work for public links
### What happened?

When not signed clicking the expand icon in the small map leads to a 403 Forbidden error instead of showing the trail.

### Version

v0.20.0

pttydou:
Fixes #1136.

Reproduction: while signed out, open a shared trail link and use the fullscreen control on the small map. The destination loses the `share` query parameter, so the fullscreen map cannot load the trail.

This preserves only the existing `share` parameter when building the fullscreen URL. Other page query state is not forwarded.

Tests:
- `node node_modules/vitest/vitest.mjs run src/lib/models/trail.test.ts --reporter=verbose`
- `node node_modules/vitest/vitest.mjs run src/lib/util/format_util.test.ts --reporter=verbose`
- `node node_modules/vitest/vitest.mjs run src/lib/util/url_util.test.ts --reporter=verbose`
- `pnpm dlx npm@11.6.0 run check`
- `pnpm dlx npm@11.6.0 run build`

I couldn't run the Docker-backed Playwright suite locally because Docker is not installed; the unit, Svelte, and production build checks above passed.

### DIFF
--- a/web/src/lib/components/trail/trail_info_panel.svelte
+++ b/web/src/lib/components/trail/trail_info_panel.svelte
@@ -71,6 +71,7 @@
     } from "$lib/stores/trail_store";
     import Combobox, { type ComboboxItem } from "../base/combobox.svelte";
     import { tags_index } from "$lib/stores/tag_store";
+    import { withShareToken } from "$lib/util/url_util";
 
     interface Props {
         initTrail: Trail;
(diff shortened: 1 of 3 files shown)


slothful-vassal:
Thank you very much for your helpful corrections.

slothful-vassal (approved):

PR #1203, opened 2026-09-05
Add a privacy policy at /privacy

Flomp:
Adds a privacy policy at `/privacy`, linked from the site footer.

Google Play requires a reachable privacy policy before the mobile app can be submitted.


slothful-vassal:
Could we clarify these points before merging?

- **Federation (section 8):** Comments on private trails can be transmitted to explicitly mentioned remote users. The receiving server may reject the activity, but the comment text has already been sent. “Only public content federates” is therefore too broad.
- **Unpublishing (section 8):** Changing a public trail or list to private currently does not emit a deletion activity. The policy should distinguish this from deleting the content.
- **Account deletion (section 11):** Existing ActivityPub snapshots containing descriptions and location data are not cleaned up by the account-deletion path. The retention wording should acknowledge these remaining snapshots.

Separately, for the planned Google Play submission, please consider documenting the actual safeguards for sensitive data during storage and transmission. Google’s [requirements](https://support.google.com/googleplay/android-developer/answer/10144311?hl=en) call for secure data-handling procedures to be covered by the policy together with any in-app disclosures.

Flomp:
Thanks, this was a good catch on all three points. While editing the policy, I actually stumbled across other issues regarding data retention upon account deletion. The PR now also contains actual code changes and has a bigger scope than initially planned.

1. Comments are now `on delete cascade` on author. Meaning, if the author goes, the comment gets deleted
2. Cascade deleted comments and summit logs now send a delete activity
3. New `util.ObjectKindFromIRI` to differentiate between object types (list, comment, trail etc.) securely
4. Broadcast account deletion to followers so other instances can also delete the user's content

### Verification
**Account deletion, single instance**

- [x] Account with at least one comment on someone else's trail deletes successfully
- [x] Account with no comments still deletes successfully
- [x] After deletion: login record, actor, trails, lists, summit logs and the account's comments are all gone
- [x] Meilisearch no longer returns the deleted account or its trails

**Two instances**

- [x] B follows A. A deletes an account. B receives `Delete(Actor)` and removes its copy of that account and their content
- [x] Same flow with A's account having public trails on B: trails disappear from B's search and feed
- [ ] A deletes a single public trail while the account still exists: B removes just that trail
- [x] Signature verification still passes for a normal `Create`/`Update`/`Announce` round trip, so the routing change did not break the happy path

slothful-vassal (changes requested):
Thanks for the update! Three issues worth addressing before merging:

1. **Deletion is announced before commit.** An API token can block account deletion and trigger a rollback, while followers have already received `Delete(Actor)`.

2. **Account deletion misses recipients who aren’t followers.**  
   Repro: User A comments on user B’s public trail on another instance. User B doesn’t follow user A. User A then deletes her account. User B received the comment, but receives neither the account deletion nor an individual comment deletion, so his copy remains.

3. **Cascaded summit-log deletions miss their audience.**  
   Repro: User A follows user B and adds a summit log to his public trail. User C, on a third instance, follows user A but not user B and receives the log. User B deletes his account. User A’s instance removes his trail and her log, but skips the log’s deletion notice because the trail is already gone. User C’s copy remains.

Flomp:
1. API Tokens are now `delete cascade`. An actor's deletion is now only announced _after_ the deletion succeeds on the local instance.
2. The audience that receives a delete announcement is now extended to:
a) accepted followers
b) actors it follows, any status
c) authors of remote trails it commented on, logged a summit on, or liked
d) remote actors who commented on, logged a summit on, or liked its trails
e) the other party of every trail or list share it is involved in
3. Summit logs now properly notify transient recipients of their deletion
4. The same error existed for comments, which is now also fixed.

slothful-vassal (changes requested):
Thanks for the updates! I found two remaining points:

1. **Account deletion still misses mentioned recipients.**

   Example: Alice comments on Bob’s local trail and mentions Carol on another Wanderer instance, with no follow or share relationship between Alice and Carol. Carol receives the comment, but falls outside the account-deletion audience listed in a–e.

   When Alice deletes her account, the individual comment deletion is skipped because its author is already gone. Carol therefore receives neither an account nor a comment deletion notice.

   Could we also include the recipients stored in previous Create/Update activities when collecting the account-deletion audience? Individual comment deletion already uses this information.

2. **The privacy policy’s activity-history disclosure needs to include comments.**

   Section 12 describes retained activity snapshots as copies of public trails, lists and summit logs. However, comments are also stored there—including comments on private trails without any remote mentions—and remain publicly accessible after deletion. The public outbox also lists these snapshots, so knowing or guessing an activity’s address is not necessary.

   This underlying behavior predates the PR, but could we update the new policy to explicitly describe comments and their public accessibility?

Flomp:
1. Mentions are now pulled from the activity log, and recipients are added to the deletion audience
2. Activities are subsequently deleted. Policy reflects this change

slothful-vassal (dismissed):

Flomp:
Fixed the scenario mentioned on discord regarding a public trail that existed before following not being deleted:
`processDeleteTrailActivity` -> `delete.go:484` -> `DeleteFromFeed` errored before deleting the trail because no feed item existed for the pre-existing trail.
This is now handled gracefully in `DeleteFromFeed`. Also added a test for good measure.

slothful-vassal (changes requested):
Hi, I'm so sorry, but I found one last issue:

- User B on instance B follows user A on instance A.
- No user on instance A follows user B.
- User B mentions user A in public trail X, which is then received by instance A.
- User B deletes trail X.

**Actual:** No Delete activity is sent to user A, even though A was mentioned in trail X and received it. The cached trail remains on instance A.

**Expected:** The deletion also reaches previous mention recipients.

Flomp:
Fixed the mention case from your last comment, then went through the whole deletion path and closed the other gaps it surfaced:

1. Mentions: trail and summit log Deletes now include the inboxes recorded in the Create/Update cc.
2. Everyone holding a copy: a trail/list Delete now also reaches remote actors who commented on, logged a summit on, liked or were shared it (including shares since revoked, read back from the recorded Announce), and actors mentioned in a local comment or summit log on the trail. Those rows cascade with the trail, so the audience is collected before deletion and announced after commit, like account deletion. A trail deleted while private still reaches those holders, but not followers; one nobody received sends nothing. cc names the followers collection plus the non-follower holders, as Create does.
3. Wrong author: a trail fetched because a summit log/comment referenced it was stored under the sender's name and rejected the real author's Delete; it's now attributed via attributedTo, accepted only from the trail's own host.
4. Foreign comments: a Delete for a non-Wanderer IRI (e.g. a Mastodon reply) is now handled as a comment Delete, mirroring the Create fallback.
5. Self-heal: refreshing a copy whose origin returns 404/410 drops the copy and returns 404 instead of "Sync failed" 500. Background sync failures are now logged. A 403 (trail taken private) deliberately keeps the copy: dropping it would cascade away comments and summit logs local users left on it.

For a future PR we should consider introducing Tombstones. E.g. if someone un-publishes a trail.

slothful-vassal (approved):

PR #1205, opened 2026-09-05
Security fixes: ActivityPub delete authorization, inbox proxy trust, collection create rules, IPv6 SSRF bypass

Flomp:
Fixes five reported security issues. One commit per advisory.

### [GHSA-xcwc-jhv2-w4h6](https://github.com/open-wanderer/wanderer/security/advisories/GHSA-xcwc-jhv2-w4h6) — ActivityPub `Delete` accepted from any remote actor

`processDeleteTrailActivity` never checked ownership (and `processDeleteListActivity` only received the actor without using it), so any remote actor could delete any local trail or list by IRI. Both handlers now compare the record's `author` against the delivering actor and reject the activity otherwise. Comment and summit-log deletes already did this.

### [GHSA-mr29-2qg5-m4fw](https://github.com/open-wanderer/wanderer/security/advisories/GHSA-mr29-2qg5-m4fw) — inbox route trusted a client-supplied `X-Forwarded-Path`

`/activitypub/activity/process` derived both the recipient lookup and the HTTP-signature verification context from `X-Forwarded-Path`, which anyone reaching the backend could set. The endpoint is now authenticated as an internal hop: the frontend proxy sends `X-Internal-Secret` (`POCKETBASE_PROXY_SECRET`), compared in constant time, and the forwarded path must additionally match the canonical inbox pattern. It fails closed when the secret is unset. The new env var is wired through both compose files, `setup.sh` (generated per install) and the env docs, and must be identical on `db` and `web`. Also drops a log line that echoed the raw `Signature` header.

### [GHSA-vj82-xj83-jp85](https://github.com/open-wanderer/wanderer/security/advisories/GHSA-vj82-xj83-jp85) — `waypoints` create rule missing the trail-ownership check

The rule was just `@request.auth.id != ""`, so any logged-in user could add waypoints to anyone's trail, attributed to whoever they liked. New migration requires the caller to own the trail (or hold an `edit` share) and binds `@request.body.author.user` to the caller.

### [GHSA-332v-mq7g-5hqf](https://github.com/open-wanderer/wanderer/security/advisories/GHSA-332v-mq7g-5hqf) — `comments`, `summit_logs`, `trail_like` did not bind the author to the caller

The trail-visibility half of each create rule was fine, but nothing tied the record's `author`/`actor` to the authenticated user, so content could be forged in another user's name. Three migrations append `@request.body.author.user = @request.auth.id` (`@request.body.actor.user` for `trail_like`); `trail_like` additionally gains the missing `@request.auth.id != ""` guard.

### [GHSA-8j93-xvwx-jjg3](https://github.com/open-wanderer/wanderer/security/advisories/GHSA-8j93-xvwx-jjg3) — SSRF filter bypassed by IPv6 transition addresses

`isPrivateOrReservedIP` and `connectorIPAllowed` only inspected the literal address, so an IPv6 form embedding an internal IPv4 (`::7f00:1`, `2002:7f00:1::`, NAT64, Teredo) passed the check and dialled the internal target. Both now run the literal address *and* any IPv4 it embeds through the same policy, unmapped, and fail closed on an address that won't parse. Adds the missing reserved prefixes (`::/96`, `64:ff9b:1::/48`, `2001::/32`, `2002::/16`, `192.88.99.0/24`) and blocks `::/96` on the safeurl client, whose own blocklist covers NAT64/6to4/Teredo but not the deprecated IPv4-compatible form.

Covered by new unit tests in `db/util/network_test.go` and `db/routes/activitypub_test.go`; `go build ./...` and `go test ./util/... ./routes/...` pass.

### Upgrade note

`POCKETBASE_PROXY_SECRET` is a required new variable. Existing installs must set the same value on both services before inbound federation will work again.


slothful-vassal (changes requested):
Thanks for the security fixes! I found two regressions:

- **[P1]** The new inbox pattern rejects existing usernames containing dots or hyphens, breaking incoming federation. Correctly signed requests returned 200 before this PR and now return 401.
- **[P2]** The new waypoint rule rejects authorized editors whenever the trail also has a view-only share. Adding a reader changes the editor’s waypoint creation response from 200 to 400.

Flomp:
Thanks. 
The ruleset problem was a good find. The same problem actually existed for list and trail updates. Hopefully fixed now.

slothful-vassal (approved):

PR #1206, opened 2026-09-06
fix: refresh trail stats when replacing GPX

### LINKED ISSUE #1199: POST /api/v1/trail/form/{id} stores a new track but keeps the old distance, duration and elevation
## What happened?

Replacing a trail's track through `POST /api/v1/trail/form/{id}` stores the new GPX and recomputes `bounding_box_diagonal`, but leaves `distance`, `duration`, `elevation_gain` and `elevation_loss` at the values derived from the **previous** track. The trail is then internally inconsistent: its bounding box describes the new track while its statistics describe the old one.

`PUT /api/v1/trail/upload` does derive all of these from the file, so the two paths disagree about who owns the numbers.

Both tracks are 11 trackpoints on a single meridian at a constant elevation, differing only in spacing: `short.gpx` steps 36.75 m every 60 s, `long.gpx` steps 665.25 m every 360 s. Uploaded on its own through `PUT /api/v1/trail/upload`, `long.gpx` gives `{"distance": 5980.53, "duration": 3600, "bounding_box_diagonal": 6645.03}`, so those are the numbers the replacement should land on.

Reproduced on a fresh trail, with the short track replaced by the long one:

```console
# create from short.gpx
$ curl -s -b cookies -X PUT -F 'file=@short.gpx' http://localhost:3000/api/v1/trail/upload
```

```json
 "elevation_gain": 0, "bounding_box_diagonal": 367.09}
```

```console
# replace with long.gpx
$ curl -s -b cookies -X POST -F "id=$ID" -F 'gpx=@long.gpx' \
    -o /dev/null -w '%{http_code}\n' http://localhost:3000/api/v1/trail/form/$ID
200
```

```json
{"gpx": "long_hwb2bqs9uh.gpx", "distance": 330.38, "duration": 600,
 "elevation_gain": 0, "bounding_box_diagonal": 6645.03}
```

The `gpx` file changed and `bounding_box_diagonal` went from 367.09 to 6645.03, matching the new track. `distance` and `duration` did not move.

Sending the statistics explicitly in the same request does update them, so the fields are writable and the client can compensate:

```console
$ curl -s -b cookies -X POST -F "id=$ID" -F 'gpx=@long.gpx' \
    -F 'distance=5980.53' -F 'duration=3600' -F 'elevation_gain=165' \
    http://localhost:3000/api/v1/trail/form/$ID
```

which answers `{"distance": 5980.53, "duration": 3600, "elevation_gain": 165, "bounding_box_diagonal": 6645.03}`.

### Expected

Either derive all statistics from the uploaded track, as `PUT /api/v1/trail/upload` does, or derive none of them and document that the caller owns them. Deriving `bounding_box_diagonal` but not the rest is the case that produces a wrong trail, because a client that does not know about the asymmetry has no reason to send the others.

If keeping the current behaviour is deliberate, so that clients can override statistics they compute better themselves, the reference should say that the fields are preserved rather than recomputed, and ideally the bounding box should follow the same rule.

## Version

## Context

Found while adding Wanderer support to Gadgetbridge, where a workout can gain a GPX track after it has already been uploaded. Gadgetbridge now sends its own distance, duration and elevation alongside the replacement track, since the watch summary is authoritative for those anyway, but it only does so because the behaviour was discovered by testing rather than from the reference.

### Version

v0.20.0

atirna:
## Why

Replacing a trail's GPX updated its geometry but kept the previous distance, duration, and elevation values.

## Verification

- Added route coverage for a GPX replacement and a metadata-only update.
- `npm run check`
- `npm run test:unit`
- Replaced a 6,004.53 m / 3,600 s trail with a 330.25 m / 600 s GPX and confirmed all four derived stats match a standalone upload.

Fixes #1199

### DIFF
--- a/web/src/lib/util/api_util.ts
+++ b/web/src/lib/util/api_util.ts
@@ -123,11 +123,11 @@ export async function uploadCreate<T>(event: RequestEvent, collection: Collectio
     return r
 }
 
-export async function uploadUpdate<T>(event: RequestEvent, collection: Collection) {
+export async function uploadUpdate<T>(event: RequestEvent, collection: Collection, formData?: FormData) {
     const searchParams = Object.fromEntries(event.url.searchParams);
     const safeSearchParams = RecordOptionsSchema.parse(searchParams);
 
-    const data = await event.request.formData();
+    const data = formData ?? await event.request.formData();
     if (!data.has('id')) {
         throw new Error("data has no id")
     }
--- a/web/src/routes/api/v1/trail/form/[id]/+server.ts
+++ b/web/src/routes/api/v1/trail/form/[id]/+server.ts
@@ -1,6 +1,8 @@
 import type { Trail } from "$lib/models/trail";
 import { Collection, handleError, uploadUpdate } from "$lib/util/api_util";
+import { fromFile, gpx2trail } from "$lib/util/gpx_util";
 import { json, type RequestEvent } from "@sveltejs/kit";
+import { ClientResponseError } from "pocketbase";
 
 /**
  * @swagger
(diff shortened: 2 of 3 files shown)


atirna:
#1228 covers this and the summit-log path too, closing mine.

PR #1222, opened 2026-09-10
fix(sharing): reject private cross-instance shares and handle mixed lists

slothful-vassal:
Sharing private trails and lists with users on another server was blocked only in the browser. Direct API requests could bypass this and expose private data on the recipient's server.

This fix also blocks these shares in the API, including changes to existing shares. Sharing a public list with another server now skips its private trails. Local recipients still receive access to the owner's private trails in that list.


Flomp:
Thanks! Verified and tested. 
Two minor nitpicks:
1. I removed the now-unused list-share-warning i18n key
2. I merged share_guard.go and share_target.go into one hook (for purely idiomatic reasons)

Flomp (approved):

PR #1226, opened 2026-09-12
fix(api): reject misnamed file uploads instead of silently no-oping

### LINKED ISSUE #1197: POST /api/v1/trail/{id}/file returns 200 but leaves the trail completely unchanged
## What happened?

`POST /api/v1/trail/{id}/file` answers `200 OK` and changes nothing at all. The stored GPX is not replaced and no derived field moves, including `bounding_box_diagonal`, which the sibling `POST /api/v1/trail/form/{id}` does update.

Because the response is a `200` carrying the trail object, a client has no way to detect that the call did nothing. I built an integration against this endpoint on the strength of the API reference, and it silently never worked.

Reproduced from scratch on a fresh trail, twice, with two clearly different tracks:

1. Create a trail from the short track:

```console
$ curl -s -b cookies -X PUT -F 'file=@short.gpx' \
    http://localhost:3000/api/v1/trail/upload
```

Trail state:

```json
 "elevation_gain": 0, "elevation_loss": 0, "bounding_box_diagonal": 367.09}
```

2. Post the much longer track to the file endpoint:

```console
$ curl -s -b cookies -X POST -F 'file=@long.gpx' \
    -o /dev/null -w '%{http_code}\n' \
    http://localhost:3000/api/v1/trail/$ID/file
200
```

3. Re-read the trail. Every field is identical, including the stored `gpx` filename:

```json
 "elevation_gain": 0, "elevation_loss": 0, "bounding_box_diagonal": 367.09}
```

For contrast, uploading `long.gpx` through `PUT /api/v1/trail/upload` yields `{"distance": 5980.53, "duration": 3600, "bounding_box_diagonal": 6645.03}`, so the two test tracks are unambiguously distinguishable and the file itself parses fine.

### Expected

Either the endpoint replaces the trail's file, as
[the API reference](https://wanderer.to/api-reference) implies by naming it "Upload trail file",
or it returns an error explaining what it wants. A `200` that means "nothing happened" is the worst of the three.

If it is meant for something other than replacing the track, the reference should say what, since `POST /api/v1/trail/form/{id}` is currently the only way I found to actually replace a trail's GPX.

## Version

## Context

Found while adding Wanderer support to Gadgetbridge, where a workout can gain a GPX track after it has already been uploaded, so the trail's track has to be replaced in place.

Possibly related, both distinct: #679 (`/upload` ignores the category) and #1025 (upload errors silent in the UI, same class of problem on a different surface).

### Version

0.20.0

Flomp:
Fixes #1197

`POST /api/v1/{trail,waypoint,user,list,summit-log}/{id}/file` forwarded the multipart body straight to PocketBase, which drops parts whose name does not match a collection field and returns the unchanged record. The API reference documented the part as `file` while the real fields are `gpx`, `photos` and `avatar`, so following the docs produced a 200 that changed nothing.

### Changes
- **trail, list, summit-log `/file`**: these duplicated `POST {entity}/form/{id}` with a strict subset of its behaviour and had no first-party callers. They are now deprecated aliases re-exporting the form handler (`deprecated: true` in the OpenAPI block, pointing to `form/{id}`), so each entity has one multipart update path. Removing the URLs is left for a breaking release.
- **waypoint, user `/file`** (no form counterpart, used by web and app): the shared `upload()` helper now takes the collection's file fields and answers `400 {"message":"missing_file","expected":[...]}` when none is present. PocketBase's `photos+`/`photos-` modifiers are accepted. OpenAPI blocks name the real fields.
- The user handler uses `handleError()` like its siblings instead of `throw error(e.status, e)`, which broke on non-HTTP errors (e.g. a Zod failure with no `status`).

### Verified
- `file=@track.gpx` (the reporter's request) → 400 listing the expected fields
- `gpx=@track.gpx` via the alias → 200, file replaced, bounding box recomputed
- user endpoint without `avatar` → 400
- existing callers (web stores, Flutter app) already send the correct names, no regression

---
Part of a stack: #1225 → #1226 → #1227 → #1228. Depends on #1225 (the alias needs `form/{id}` to honour the path id).


slothful-vassal (changes requested):
1. One regression I noticed: `waypoints_update()` sends a `/file` request even when no photos were added or removed. Renaming or moving a waypoint therefore now triggers `400 missing_file`. The waypoint metadata has already been saved at that point, but the remaining trail update is aborted.

   Could we skip the upload request when there are no photo changes? A regression test for this case would also be helpful.

2. Is #1197 actually fixed by this change? The deprecated aliases delegate to the form handlers, which call `uploadUpdate()`, while the new file-field validation only exists in `upload()`. Wouldn’t the reporter’s `file=@track.gpx` request therefore still reach PocketBase unchanged and return `200` without replacing the GPX?

Flomp:
1. waypoints_update() now shortcuts if there are no changes to photos/gpx and never calls the file endpoint
2. The /file endpoints now assert that there is a file field present and errors otherwise

slothful-vassal (approved):

PR #1227, opened 2026-09-12
fix(api): answer unknown /api paths and wrong methods with JSON

### LINKED ISSUE #1196: Unknown /api/v1/* paths return the HTML app shell instead of a JSON error
## What happened?

A request to a path under `/api/v1/` that does not exist returns the SvelteKit application's HTML document. The status code is correct (`404`), but the body is a full HTML page:

```console
$ curl -s -w '\n%{http_code}\n' http://localhost:3000/api/v1/nonexistent_endpoint_xyz
<!doctype html>
<html lang="en">
...
404
```

An API client parsing the response as JSON gets a syntax error pointing at `<`, rather than a usable error. That turns "you called the wrong path" into a confusing parse failure, and it makes a typo'd endpoint look like a server or serialization problem.

I hit this while probing which auth endpoints exist: `POST /api/v1/auth/register` returns the app shell, which reads as "the endpoint is there but returned something strange" instead of "there is no such endpoint".

### Expected

Anything under `/api/v1/` should return a JSON body on error, for example `{"message": "not found"}`, matching the shape the API already uses elsewhere (`{"message": "data has no id"}`). The HTML fallback should apply to page routes only.

## Version

0.20.0 (`flomp/wanderer-web@sha256:d76177573caab135e8167179c9954168b5c582a48cbef0fcb7e1571af0c84d98`), stock `docker-compose.yml` with PocketBase and Meilisearch.

## Context

Found while adding Wanderer support to Gadgetbridge

### Version

v0.20.0

Flomp:
Fixes #1196

An unmatched path under `/api` rendered the SvelteKit HTML error page, so an API client hitting a typo'd or non-existent endpoint got a document starting with `<` instead of a parseable error. `Accept: application/json` made no difference. A wrong method on an existing route returned SvelteKit's `text/plain` 405.

### Changes
`hooks.server.ts` gets a new handle, first in the `sequence`:
- no route matched under `/api/` → `404 {"message":"not_found"}`, returned before CSRF and auth run (no PocketBase round-trips for garbage paths)
- 405 from an existing route → `{"message":"method_not_allowed"}`, `Allow` header preserved
- the CSRF rejection answers JSON for `/api/` paths regardless of the `Accept` header

Page routes keep the HTML fallback.

Also in this PR, same theme of API clients getting a usable error: unauthenticated `PUT /api/v1/trail/upload` crashed on `event.locals.settings` and returned `500 "Cannot read properties of undefined (reading 'privacy')"`. The route isn't among the ones the auth hook guards, so it now checks `locals.user` itself and answers `401 Unauthorized`, like `/trail/download` does.

### Verified
- `GET /api/v1/nonexistent`, `POST /api/v1/auth/register`, `/api/v1/trail/x/nope` → JSON 404
- `DELETE /api/v1/trail/upload` → JSON 405 with `allow: PUT`
- `/nonexistent-page` → HTML error page as before; cross-site form POST to `/login` → text 403 as before
- existing API and page routes unchanged
- unauthenticated `PUT /api/v1/trail/upload` → `401 {"message":"Unauthorized"}`

---
Part of a stack: #1225 → #1226 → #1227 → #1228. Independent of the layers below; stacked for a single review pass.


slothful-vassal (changes requested):
One small issue: replacing the 405 response preserves only `Allow`, dropping the `Set-Cookie` headers already added by the auth hook. This includes the refreshed `pb_auth` cookie and any updates or deletions of `meilisearch_token`.

Could we preserve the existing response headers and adjust the body-related headers for JSON? A regression test checking that cookies survive the conversion would also be helpful.

Flomp:
Headers are now retained through error responses.

slothful-vassal (approved):

PR #1228, opened 2026-09-12
fix(api): derive trail and summit log statistics when a track is replaced

### LINKED ISSUE #1199: POST /api/v1/trail/form/{id} stores a new track but keeps the old distance, duration and elevation

Flomp:
Fixes #1199

Replacing a trail's track through `POST /api/v1/trail/form/{id}` stored the new file and let the Go hook recompute `polyline` and `bounding_box_diagonal`, but `distance`, `duration`, `elevation_gain` and `elevation_loss` kept the values of the previous track. Those are only derived by `PUT /api/v1/trail/upload`, so the trail ended up internally inconsistent.

### Changes
`form/{id}` now runs a `gpx` part through the same pipeline as `/upload` (`applyGpxToTrailForm` in `gpx_util.ts`):
- the file is converted to GPX first (FIT, KML, KMZ, TCX accepted), so what is stored is what the Go hook can parse
- `distance`, `duration`, `elevation_gain`, `elevation_loss`, `lat`, `lon` are derived from the new track for every field the request does **not** set explicitly. Fields sent by the client always win, so the web UI (which sends its own numbers) and clients with better measurements are unaffected.
- elevation is corrected against the elevation service as on `/upload`; failure falls back to the file's elevation with a warning
- unparseable file → `400 Invalid file`, same as `/upload`
- the deprecated `/file` alias inherits the behaviour

Not touched on a replace: `name`, `description`, `date`, waypoints.

`POST /api/v1/summit-log/form/{id}` had the same gap and gets the same treatment (second commit): the helper takes the list of fields to derive, and summit logs derive `distance`, `duration`, `elevation_gain`, `elevation_loss` — matching what the summit log modal computes client-side.

### Verified
- gpx only → all six fields derived, bbox recomputed
- gpx + explicit `distance`/`duration`/`elevation_gain` → explicit values kept, unsent `elevation_loss` derived
- name-only update → statistics untouched
- KML in → stored as `.gpx` containing GPX, parsed by the Go hook
- garbage file → 400
- summit log: gpx only → four stats derived; gpx + explicit `distance` → explicit kept, rest derived; text-only update → stats untouched

---
Part of a stack: #1225 → #1226 → #1227 → #1228. This is the top layer.


slothful-vassal (changes requested):
I found a few edge cases that would be worth addressing before merging:

1. **Route-only GPX files:** The metrics calculation only handles `<trk>`, so a valid file using `<rte>`/`<rtept>` produces zero distance, duration and elevation totals. The new helper then saves those zeros for both trails and summit logs. Could we support route points here, or avoid overwriting statistics when they cannot be calculated?

2. **GPX namespace prefixes:** Valid files using `<g:gpx>` with the GPX namespace were accepted by the existing Go parser, but the new parser rejects them with `400 Invalid file` because it expects `xml.gpx`. Could we preserve compatibility with these files, including when the client supplies all statistics explicitly?

3. **Conversion errors:** `fromFile()` runs outside the `try/catch`, so malformed KMZ, KML or TCX files can return `500` instead of the documented `400`. Could we include conversion in the same error handler as GPX parsing?

4. **Validation order:** On public instances, anonymous requests now reach archive decompression, XML parsing and elevation lookup before PocketBase rejects the update. Invalid or mismatched IDs also trigger this work. Could we add the same early user check as `/trail/upload` and validate IDs before processing the file?

Regression tests for these cases would help protect the new behavior.

Flomp:
1. I think this would be a follow-up PR. The issue exists throughout the app. Also, when a trail is first uploaded. For now, we prevent an empty trk point list from overwriting the pre-existing elevation gain/loss, distance and duration values on a trail
2. Added `tagNameProcessors` stripping the prefix in gpx.ts
3. Moved `fromFile()` inside `try/catch`
4. Added two guards to the `POST` routes that return early if there is no authorised user

slothful-vassal (changes requested):
Thanks for the fixes. I found one side effect of the namespace change: `tagNameProcessors` also strips prefixes from Garmin extensions. When `trail2gpx()` exports the file, heart-rate and temperature elements end up in the GPX namespace instead of the Garmin namespace.

I reproduced this with the real Garmin Connect file added in [GPSBabel PR #990](https://github.com/GPSBabel/gpsbabel/pull/990): all 68 heart-rate and 68 temperature values lose their Garmin namespace during export. This also affects previously imported tracks.

Could we move the replacement-statistics calculation to the existing Go backend? We already use `gpxgo` there, and `metricsFromGPX()` in `db/plugins/importer/importer.go` derives distance, duration, elevation gain/loss and the start position.

My suggestion would be to reuse that calculation for trail and summit-log replacements, while:

- keeping explicitly supplied client values authoritative;
- preserving the uploaded/converted GPX file unchanged during metric extraction;
- removing the global prefix-stripping change from the web parser.

This would reuse our existing GPX processing and avoid adding another namespace-normalization layer to maintain.

Flomp:
In an ideal world, I would move all metric calculations to the Go backend. Right now, that’s more or less done on the client side. The app and the web client are both doing their own thing. That would also resolve your remark. However, I see that as a separate (major) PR. For example, there’s currently no VALHALLA_URL in the Go backend. 

I commited a more targeted fix now that only strips GPX namespace attributes and leaves all others alone.

slothful-vassal (approved):

PR #1238, opened 2026-09-15
fix(security): guard anonymous collection write rules

Flomp:
Federated trails, lists, and comments could be changed or deleted without logging in. The write rules compare the record owner's local user ID with the requesting user's ID, but accounts from other instances have no local user ID and anonymous requests have none either, so the comparison matched. The same applied to creating or removing shares, likes, and follows on behalf of remote accounts. Follow-up to #1232, which fixed the read rules.

The migration requires authentication before checking ownership in every create, update, and delete rule. Owners and edit-share recipients keep their existing permissions.


slothful-vassal (approved):
Thanks for following up on #1232! I made a small correction to the tests: `nil` rules mean superuser-only access, so they’re now checked instead of skipped. I also adjusted the expected owner permissions for each collection.

All permission tests pass, and I verified that accidentally removing owner or editor access now makes them fail as expected.

PR #1260, opened 2026-09-21
fix: use local calendar days for trail date filters

slothful-vassal:
Date filters currently interpret selected dates as UTC midnight and use the beginning of the end date as the upper bound. This shifts the range relative to the displayed local dates and excludes trails recorded later on the selected end date.

This PR treats selected dates as full calendar days in the browser’s local time zone, matching the existing date display:

- Include the start of the selected first day.
- Include the entire selected last day by using the start of the following day as an exclusive upper bound.
- Calculate each boundary independently, correctly handling daylight saving transitions without assuming every day lasts 24 hours.
- Ignore invalid date inputs while retaining any valid boundary.

The same behavior applies to the trail overview and map search. For example, selecting September 7 as both boundaries now includes all trails dated September 7 in the browser’s time zone.


Flomp:
Thanks for the thorough tests, but I don't think local-time bounds are the right model here. Trail date is set via `<input type="date">`, so a trail entered as `2026-09-07` is stored and indexed as UTC midnight (`2026-09-07T00:00:00Z`, see `db/util/meilisearch.go:89`), and the API truncates it back to the UTC calendar day on every response (`web/src/routes/api/v1/trail/+server.ts:125`). With this change, a user in e.g. America/Los_Angeles who filters `09-07..09-07` gets a window of `09-07T07:00Z–09-08T07:00Z` and no longer finds the trail they entered as `09-07` -> regression from the current behaviour. It also diverges from the summit log and profile statistics filters, which compare against UTC calendar days (`summit_log_store.ts:152-157`, `profile_statistics.ts:145-149`).

The "matches the existing display" argument holds only because the cards do `new Date("2026-09-07").toLocaleDateString()`, which renders that LA trail as `9/6`. That's a display bug, not something the filter should align with. I think the actual defect is just the inclusive `<=` on the end bound dropping GPX-timestamped trails later on the end date. I'd suggest keeping UTC calendar days and making the end exclusive (`date < Date.parse(endDate + "T00:00:00Z") / 1000 + 86400`), while keeping your invalid-input handling.

slothful-vassal:
Thanks for catching this — you’re right. Updated the filter to use UTC calendar days with an exclusive next-day end bound.

Flomp (approved):

PR #1268, opened 2026-09-23
perf(images): use pocketbase thumbnails for cards, galleries, and avatars

fparadise:
### Summary

Second PR regarding perf optimizations. This one is about reducing network load. 

Use PocketBase image thumbnails across cards, gallery previews, map popups, and avatars instead of loading full-resolution original files.

### Context & Problem

Previously, views across Wanderer fetched original uncompressed photos even when rendered at small thumbnail sizes (e.g. 40px avatars, 200px gallery grids, or 600px cards). For a trail with several photos, opening the trail page could download tens of megabytes of images on initial load.

PocketBase supports on-the-fly thumbnail generation via `?thumb=WxH` (e.g., `600x0`, `100x100`), and `getFileURL` already supports an optional `thumb` parameter.

### Changes

- **Core**: Guard `getFileURL` so that `?thumb=` is only appended to images, not video files (`isVideoURL`).
- **Trail View & Map View (`trail_info_panel.svelte`)**:
  - Request `600x0` thumbnails for header banner images (`getHeaderPhotos`).
  - Request `600x0` thumbnails for the photos tab gallery grid (PhotoSwipe lightbox retains full-resolution originals on click/zoom).
  - Request `100x100` thumbnails for comment author avatars.
- **Lists (`list_card.svelte`, `list_panel.svelte`, `list_search_modal.svelte`, `lists/edit`)**:
  - Request `100x100` / `300x300` for list avatars and `600x0` for trail thumbnails.
- **Trail Edit (`photo_picker.svelte`)**:
  - Request `600x0` thumbnails in the photo reorder grid.
- **Map Popups (`maplibre_util.ts`)**:
  - Request `600x0` thumbnail in `createPopupFromTrail`.
- **Waypoints & Timeline (`trail_timeline.svelte`, `waypoint_card.svelte`)**:
  - Request `600x0` / `300x300` for previews (full resolution kept for lightbox).
- **Social / Feed (`activity_card.svelte`, `feed_card.svelte`, `nav_bar.svelte`)**:
  - Request `600x0` for activities and `100x100` for navbar avatar.
- **Tests**: Added unit tests for `getFileURL` in `file_util.test.ts`.

### Benchmark / Network Payload

Measured on a sample trail photo:
- Original file: **2,159 KB**
- Thumbnail (`?thumb=600x0`): **83 KB** (**-96% payload reduction**)
- Avatar (`?thumb=100x100`): **4.4 KB** (**-99% payload reduction**)

### Verification

- `npx vitest run`: 20/20 test files passed (160 tests passing).
- `npm run check`: 0 errors, 0 warnings.
- Verified that PhotoSwipe lightbox and ZIP exports still use full-resolution original files.


Flomp:
Thanks for this! I changed 2 things:

**`fix(db): register the thumbnail sizes the web UI requests`** 
PocketBase only honors a `?thumb=` size when it's the built-in `100x100` default or is listed in that file field's `thumbs` option; every other size silently falls back to the original file with a 200. Only `trails.photos` had `600x0` registered, so about half the call sites here were no-ops — `lists.avatar`, `users.avatar`, `waypoints.photos` and `summit_logs.photos` were all `thumbs: []`. The migration registers the sizes you use, which is what makes the rest of the PR actually take effect. I also added `db/tests/thumbs/thumb_sizes_test.go`: a table of every size the UI requests, so this can't silently drift again. With the migration removed it fails on exactly the seven sizes that were dead, naming the component for each.

**`fix(web): detect video files by extension instead of substring`** 
`url.includes("ogg")` is true for `doggo.jpg`, so a photo with such a name would quietly skip its thumbnail. It now matches the extension with any query string stripped, which also handles uppercase extensions. I checked all the `isVideoURL` call sites first, since some pass bare filenames and some pass built URLs; both work, and blob URLs still return false as before.

Flomp (approved):

